PR #57, opened 2026-04-17
fix: resolve system default input device explicitly for Bluetooth HFP

dmetzler:
## Summary

- **Bug**: AirPods (and other Bluetooth devices) produce no audio when input is set to "System Default"
- **Root cause**: When `selectedDeviceID` is `nil`, `startCapture()` skipped both the explicit `AudioUnitSetProperty` call and the `waitForBluetoothHFP()` settling delay — AirPods never switched from A2DP to HFP/SCO mic profile
- **Fix**: Resolve the system default device ID via `kAudioHardwarePropertyDefaultInputDevice` and route it through the same per-engine device setup path as explicitly-selected devices

## What changed

`AudioEngine.startCapture()` now uses `selectedDeviceID ?? getDefaultInputDeviceID()` so that "System Default" gets:
1. Explicit `AudioUnitSetProperty` on the AudioUnit (no stale cache)
2. Bluetooth transport type detection → `waitForBluetoothHFP()` settling
3. Tap format built from actual hardware sample rate and channel count

Net diff: **+13 / -12 lines** in `AudioEngine.swift`. No new files, no API changes.

### DIFF
--- a/wispr/Services/AudioEngine.swift
+++ b/wispr/Services/AudioEngine.swift
@@ -97,13 +97,14 @@ actor AudioEngine {
         self.engine = audioEngine
         let inputNode = audioEngine.inputNode
 
-        // Assign the selected device to this engine's input AudioUnit
-        // BEFORE reading any format or installing taps (Req 2.1, 2.3).
-        // For selected devices, build an explicit tap format from CoreAudio
-        // since inputNode's cached format may be stale after device switches.
-        // For system-default devices, nil lets AVAudioEngine use its cache.
+        // Assign the input device to this engine's AudioUnit BEFORE reading
+        // any format or installing taps (Req 2.1, 2.3).
+        // Always resolve the actual device — even for "System Default" — so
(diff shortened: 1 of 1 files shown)


sebsto:
Hey @dmetzler What is the bug this PR fixes ? How can I reproduce the bug to verify this change ?
Here is my test procedure 
- start wispr
- verify sound input is "System"
- press the hotkey and try a dictation 
- connect the airpods
- verify in wispr settings that the input is still system default 
- start the dictation again 

It works.

The Bluetooth HFP bug has been fixed in 
https://github.com/sebsto/wispr/pull/57 

Is this different ?

dmetzler:
So, I was able to reproduce the bug today (2026/05/19) on 1.9.2.  I wanted to do a recording using QuickTime. But when I recorded, it finally worked. So there must be something weird in my audio input precedence configuration.
Let's call this "do not reproduce". If I'm able to reproduce it constantly, I'll reopen an issue.
Sorry for the noise!

PR #61, opened 2026-04-29
 feature: meeting transcriber

gbrunoo:
adding the possibility to record meetings by recording from own microphone and from the computer audio stream. the two transcript stay separate and can then subsequently copied elsewhere


.kiro/.DS_Store
sebsto, line 1: Can you remove this file ?

build/.DS_Store
sebsto, line 1: Can you remove this file

.kiro/specs/meeting-transcription/design.md
sebsto, line 1: Can you rename that file this file. It's not a todo, it's a design for the new feature and should be with other design documents

wispr.xcodeproj/project.pbxproj
sebsto, line 1: This file should not be committed to the project.

wispr/UI/Meeting/MeetingWindowPanel.swift
```diff
@@ -0,0 +1,120 @@
...
+        guard let panel, isVisible else { return }
+        panel.orderOut(nil)
+        isVisible = false
+    }
+
+    // MARK: - NSWindowDelegate
+
+    /// Called when the user closes the window via the red X button.
+    /// Syncs `isWindowVisible` back to false so the menu item can reopen it.
+    nonisolated func windowWillClose(_ notification: Notification) {
```
sebsto: The default SwiftUI views are isolated to MainActor. Why declare it nonisolated and the run the code in a MainActor isolated block ?  I may miss something, otherwise this can be simplified.

wispr/UI/Onboarding/OnboardingFlow.swift
sebsto, line 1: I don't understand why adding / heading steps in the onboarding flows. 
I think I understand you changed the app to work in a different mode for meeting transcription. Either teh app install itself as a meeting transcription either as dictation. If this is the case, this must be changed :-) The app should stay polyvalent and work with two different activations mode : transcript or dictation and user must be able to use one or the other easily (typically, two different actions) - not a persistent mode.

IMHO, the onboarding flow must not be modified to support this new capability

wispr/Services/SettingsStore.swift
```diff
@@ -118,6 +118,12 @@ final class SettingsStore {
         }
     }
 
+    // MARK: - App Variant
+
+    /// When true, the app is running as the meeting transcription variant.
+    /// Disables features that require accessibility permission (text insertion at cursor).
+    var isMeetingMode: Bool = false
```
sebsto: As described in my previous comment. OK to work in meeting mode temporaly. If this is needed we can leave it as-is. But onboarding can not be touched by this

wispr/Utilities/ModelPaths.swift
sebsto, line 1: Why the changes in this file ?

.gitignore
sebsto, line 1: This file should not be part of the commit

sebsto (changes requested):
Hello, @gbrunoo Thank you so much for this submission !  This will add a much demanded capability to the app. 

I left a few comments
- files that should not be part of the commit
- questions about the mode operation that influence the onboarding
- IMHO, the onboarding flow should not be touched by this change 

Also, there is no test for the new audio engine and state manager.  Can you add some to verify the logic / state of these two new components ?

Please answer on the comments. I will pull these changes in and compile / test the next version.

sebsto:
Hello @gbrunoo Thank you fo rthe new commits. Now, there 128 files changed in the project.
This is too big to review.  Looks like most file haven't really be touched, just the permission.
Can you revert that and only send the files that have actually be changed ?
Thank you

.kiro/specs/meeting-transcription/design.md
sebsto: @gbrunoo This file is still incorrectly named.  It should not be named `todo.md` and must reside in the specs directory, not in tasks.

sebsto (changes requested):
Thank you for having reverted the multi mode aspects. Looks liek your force push changed the permission on all projects files.  This PR now touches 128 files, most of them with a permission change. Can you revert this to only send the files that have actually be changed ?
also, the `task/toto.md` must be renamed and placed into `.kiro/specs` directory.
Thank you

wispr/UI/Onboarding/OnboardingFlow.swift
gbrunoo: The meeting transcriber requires additional rights as the app can record the system sounds. If not included on the onboarding the user will have to manually add the rights on the first attempt to use the meeting transcriber

sebsto:
Thank you @gbrunoo This is much easier to read.
Look great in general, but with a lot of points that are non compliant with Swift 6 structures concurrency (`@unchecked Sendable`, unmanaged `Tasks`, `nonisolated(unsafe)` etc)

Use the Kiro power skill to direct AI Coding agent towards the correct set of rules 
https://github.com/sebsto/swift-kiro-power

wispr/Services/MeetingAudioEngine.swift
```diff
@@ -0,0 +1,356 @@
...
+        if let stream = systemStream {
+            try? await stream.stopCapture()
+        }
+    }
+}
+
+// MARK: - ScreenCaptureKit Audio Output Handler
+
+/// Receives audio sample buffers from SCStream and converts them to Float32 arrays.
+final class SystemAudioOutputHandler: NSObject, SCStreamOutput, @unchecked Sendable {
```
sebsto:
```swift
final class SystemAudioOutputHandler: NSObject, SCStreamOutput, @unchecked Sendable {
    private let onSamples: @Sendable ([Float]) -> Void
```

The class is `final`, its sole stored property is an immutable `let` of `@Sendable` closure type, and `NSObject` is already `@unchecked Sendable` in the SDK. The compiler can verify `Sendable` conformance directly — `@unchecked` is not needed and bypasses that verification.

**Severity:** Low
**Fix:** Replace `@unchecked Sendable` with plain `Sendable`.

wispr/Services/MeetingStateManager.swift
```diff
@@ -0,0 +1,311 @@
...
+/// Orchestrates microphone capture, runs continuous chunked transcription,
+/// and assembles a timestamped transcript.
+///
+/// Note: Currently captures microphone only. System audio capture (for remote
+/// meeting participants) requires Screen Recording permission which is
+/// incompatible with App Sandbox. Speaker labels default to "You" for all
+/// entries until system audio support is added.
+@MainActor
+@Observable
+final class MeetingStateManager {
```
sebsto, line 33: `MeetingStateManager` is `@MainActor`. Tasks spawned from its methods inherit MainActor isolation. The `await MainActor.run { }` wrappers inside those tasks are redundant hops that add overhead and mislead readers into thinking the code is off the main actor.

| Location | Code |
|----------|------|
| `startMicTranscription()` | `await MainActor.run { self.transcript.entries.append(...) }` |
| `startSystemTranscription()` | `await MainActor.run { self.transcript.entries.append(...) }` |
| `startMicLevelConsumption()` | `await MainActor.run { self?.micLevel = level }` |
| `startSystemLevelConsumption()` | `await MainActor.run { self?.systemLevel = level }` |
| `startTimer()` | `await MainActor.run { self?.elapsedTime = ... }` |

**Severity:** Low
**Fix:** Remove all five `await MainActor.run { }` wrappers; access properties directly.

sebsto, line 33: The manager spawns five unstructured `Task { }` stored in properties and manually cancelled in `stopMeeting()`:

| Property | Spawned in |
|----------|------------|
| `micTranscriptionTask` | `startMicTranscription()` |
| `systemTranscriptionTask` | `startSystemTranscription()` |
| `micLevelTask` | `startMicLevelConsumption(_:)` |
| `systemLevelTask` | `startSystemLevelConsumption(_:)` |
| `timerTask` | `startTimer()` |

If `stopMeeting()` is never called (e.g. the manager is deallocated while recording), the tasks leak and keep running. There is no `deinit` safety net.

**Severity:** Medium
**Fix:** Replace the five independent tasks with a single `withTaskGroup` launched from `startMeeting()`. The group naturally cancels all child tasks when the parent scope exits. Alternatively, store a single parent `Task` that spawns child tasks via `async let` or a task group, so cancelling the parent cascades to all children.

wispr/Services/MeetingAudioEngine.swift
```diff
@@ -0,0 +1,356 @@
...
+import os
+
+/// Actor responsible for capturing both microphone and system audio simultaneously.
+///
+/// Uses `AVAudioEngine` for microphone input and `SCStream` (ScreenCaptureKit)
+/// for system audio capture. Both streams are resampled to 16kHz mono Float32.
+///
+/// If system audio capture fails (e.g. permission denied, sandbox restriction),
+/// the engine continues with mic-only capture and logs a warning.
+actor MeetingAudioEngine {
```
sebsto, line 28:
```swift
// In installTap closure
Task {
    await self.processMicSamples(samples)
}

// In SystemAudioOutputHandler callback
Task {
    await self.processSystemSamples(samples)
}
```

Each audio buffer spawns a new unstructured `Task` to hop into the actor's isolation domain. Under heavy audio load this creates many short-lived tasks with no backpressure.

**Severity:** Low-Medium
**Fix:** Use an `AsyncStream` as the bridge instead — the tap yields samples into a continuation, and a single structured task consumes them inside the actor. This gives natural backpressure and one task instead of hundreds.

wispr/wisprApp.swift
```diff
@@ -255,13 +286,17 @@ final class WisprAppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate
 
     func applicationWillTerminate(_ notification: Notification) {
         overlayObservationTask?.cancel()
+        meetingObservationTask?.cancel()
         hotkeyObservationTask?.cancel()
         permissionMonitoringTask?.cancel()
         updateCheckTask?.cancel()
 
+        // Stop any active meeting session
```
sebsto:
```swift
if let msm = meetingStateManager {
    Task { await msm.stopMeeting() }
}
```

This spawns an unstructured task during app termination. The process may exit before the task completes, meaning `stopMeeting()` (which flushes buffers, cancels child tasks, stops audio capture) may never finish.

**Severity:** Medium
**Fix:** `await` the stop synchronously (acceptable at termination)

sebsto (changes requested):
Much better now - still a couple of lines that are not compliant with swift 6 structred concurrency rules

gbrunoo:
Hi @sebsto, thank you for the thorough review! I've addressed all the feedback:

Swift 6 Structured Concurrency fixes:

    MeetingAudioEngine: Replaced per-buffer Task spawning with AsyncStream bridge pattern — tap callbacks yield into continuations, single consumer tasks read on the actor. Eliminates hundreds of short-lived tasks and provides natural backpressure. Also replaced @unchecked Sendable with plain Sendable on SystemAudioOutputHandler.
    MeetingStateManager: Removed all 5 redundant await MainActor.run {} wrappers (Tasks inherit @MainActor isolation). Replaced 5 unstructured Task properties with a single recordingTask using withTaskGroup — cancelling the parent cascades to all children. Added cancelRecording() for synchronous cleanup.
    wisprApp.swift: Replaced fire-and-forget Task { await msm.stopMeeting() } in applicationWillTerminate with synchronous cancelRecording().

Ready for re-review when you have a chance!

PR #63, opened 2026-05-04
Meeting transcriber

gbrunoo:
Hi @sebsto, thank you for the thorough review! I've addressed all the feedback:

**Swift 6 Structured Concurrency fixes:**
- **`MeetingAudioEngine`**: Replaced per-buffer `Task` spawning with `AsyncStream` bridge pattern — tap callbacks yield into continuations, single consumer tasks read on the actor. Eliminates hundreds of short-lived tasks and provides natural backpressure. Also replaced `@unchecked Sendable` with plain `Sendable` on `SystemAudioOutputHandler`.
- **`MeetingStateManager`**: Removed all 5 redundant `await MainActor.run {}` wrappers (Tasks inherit `@MainActor` isolation). Replaced 5 unstructured `Task` properties with a single `recordingTask` using `withTaskGroup` — cancelling the parent cascades to all children. Added `cancelRecording()` for synchronous cleanup.
- **`wisprApp.swift`**: Replaced fire-and-forget `Task { await msm.stopMeeting() }` in `applicationWillTerminate` with synchronous `cancelRecording()`.

Ready for re-review when you have a chance!


wispr/Services/MeetingStateManager.swift
```diff
@@ -0,0 +1,285 @@
...
+        pasteboard.clearContents()
+        pasteboard.setString(text, forType: .string)
+    }
+
+    /// Saves the transcript to a text file via save panel.
+    func exportTranscript() {
+        let text = transcript.asPlainText()
+        guard !text.isEmpty else { return }
+
+        let panel = NSSavePanel()
```
sebsto: When clicking on "export" on the UI, The app crashes with this message 
"Unable to display save panel: your app has the User Selected File Read entitlement but it needs User Selected File Read/Write to display save panels. Please ensure that your app's target capabilities include the proper entitlements."

wispr/Services/MeetingAudioEngine.swift
```diff
@@ -0,0 +1,407 @@
...
+        let status = CMBlockBufferGetDataPointer(
+            blockBuffer, atOffset: 0, lengthAtOffsetOut: nil, totalLengthOut: &length,
+            dataPointerOut: &dataPointer)
+
+        guard status == noErr, let data = dataPointer, length > 0 else { return }
+
+        let floatCount = length / MemoryLayout<Float>.size
+        guard floatCount > 0 else { return }
+
+        let samples = Array(
```
sebsto:
```
  let samples = Array(                                                                                                                                               
      UnsafeBufferPointer(                                                                                                                                           
          start: data.withMemoryRebound(to: Float.self, capacity: floatCount) { $0 },                                                                                
          count: floatCount                                                                                                                                          
      ))                                                                                                                                                             
```
                                                                                                                                                                     
  `withMemoryRebound` returns a pointer that is only valid inside the closure. Returning `$0` and using it outside is undefined behavior — the pointer may be invalidated
   by the time `Array.init` reads from it.

This should be:

```                                                                                                                                                                     
  let samples = data.withMemoryRebound(to: Float.self, capacity: floatCount) { ptr in
      Array(UnsafeBufferPointer(start: ptr, count: floatCount))                                                                                                      
  } 
```

wispr/Services/MeetingStateManager.swift
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    // MARK: - Timer
+
+    private func runTimer() async {
+        while !Task.isCancelled {
+            try? await Task.sleep(for: .seconds(1))
+            guard !Task.isCancelled else { break }
+            self.elapsedTime = self.transcript.formattedDuration ?? "0:00"
```
sebsto:
```
self.elapsedTime = self.transcript.formattedDuration ?? "0:00"                                                                                                     
```
                                                                                                                                                                     
  `MeetingTranscript.formattedDuration` returns `String` (non-optional). The `??` is unnecessary and will generate a compiler warning. Should be:                          
                   
```                                                                                                                                                  
  self.elapsedTime = self.transcript.formattedDuration  
```

wispr 19.06.04/UI/Meeting/MeetingTranscriptView.swift
```diff
@@ -0,0 +1,235 @@
...
+
+/// The main content view for the meeting transcription window.
+///
+/// Shows recording controls at the top, a scrolling transcript in the middle,
+/// and export actions at the bottom.
+struct MeetingTranscriptView: View {
+    @Environment(MeetingStateManager.self) private var meetingState: MeetingStateManager
+    @Environment(UIThemeEngine.self) private var theme: UIThemeEngine
+
+    @State private var scrollProxy: ScrollViewProxy?
```
sebsto: This is declared but never written to. The `ScrollViewReader` proxy is used inline. Remove the dead state.

wispr 19.06.04/Models/MeetingTranscript.swift
```diff
@@ -0,0 +1,63 @@
...
+struct MeetingTranscript: Sendable, Equatable {
+    var entries: [MeetingTranscriptEntry] = []
+    let startTime: Date
+
+    init(startTime: Date = Date()) {
+        self.startTime = startTime
+    }
+
+    /// Formats the entire transcript as plain text for export.
+    func asPlainText() -> String {
```
sebsto: A new `DateFormatter` is allocated for every call, and the method iterates all entries. For a long meeting (hundreds of entries), this is wasteful. Consider a static
   cached formatter or pass one in:

```                                                                                                                                                                     
  func asPlainText() -> String {
      let formatter = DateFormatter()                                                                                                                                
      formatter.dateFormat = "HH:mm:ss"                                                                                                                              
      return entries.map { ... }.joined(separator: "\n")                                                                                                             
  } 
```                                                                                                                                                                 
                                                                                                                                                                     
  Same issue in `MeetingTranscriptView.formatTime(_:)` — called per row in a LazyVStack.  

Also probably an occasion to not repeat yourself and have only one formatter utility method

Sources/WisprApp/Services/MeetingAudioEngine.swift
```diff
@@ -0,0 +1,407 @@
...
+
+        let audioEngine = AVAudioEngine()
+        self.micEngine = audioEngine
+        let inputNode = audioEngine.inputNode
+
+        nonisolated(unsafe) var converter: AVAudioConverter?
+        nonisolated(unsafe) var sampleRateRatio: Double = 0
+
+        let bridgeContinuation = micBridgeCont
+        inputNode.installTap(onBus: 0, bufferSize: 4096, format: nil) { [weak self] buffer, _ in
```
sebsto:
```  
inputNode.installTap(...) { [weak self] buffer, _ in                                                                                                               
      guard self != nil, buffer.frameLength > 0 else { return }                                                                                                      
      // self is never used after this point                                                                                                                         
      ...                                                                                                                                                            
      bridgeContinuation.yield(samples)                                                                                                                              
  }                                                                                                                                                                  
```
                  
  The `weak self` check serves no purpose since self isn't used in the closure body (the continuation is captured directly). Either remove the capture or use self meaningfully (e.g., to check `isCapturing`).

wispr 19.06.04/Services/PermissionManager.swift
```diff
@@ -86,7 +87,24 @@ final class PermissionManager {
...
+
+    /// Opens System Settings to the Screen Recording privacy pane.
+    /// Required for meeting mode system audio capture via ScreenCaptureKit.
+    func openScreenRecordingSettings() {
+        guard let url = URL(string: "x-apple.systempreferences:com.apple.preference.security?Privacy_ScreenCapture") else { return }
+        NSWorkspace.shared.open(url)
+    }
+
+    /// Checks if Screen Recording permission is available by attempting to enumerate shareable content.
+    func checkScreenRecordingPermission() async -> Bool {
```
sebsto: The method exists but nothing in this PR calls it before attempting system audio capture. Either wire it into the meeting start flow (show UI guidance before the SCK prompt) or remove it to avoid dead code.

wispr 19.06.04/Services/PermissionManager.swift
```diff
@@ -86,7 +87,24 @@ final class PermissionManager {
         guard let url = URL(string: "x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone") else { return }
         NSWorkspace.shared.open(url)
     }
-    
+
+    /// Opens System Settings to the Screen Recording privacy pane.
+    /// Required for meeting mode system audio capture via ScreenCaptureKit.
+    func openScreenRecordingSettings() {
```
sebsto: do we need this ?

wispr 19.06.04/Services/PermissionManager.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 import AVFAudio
 import ApplicationServices
 import AppKit
+import ScreenCaptureKit
```
sebsto: if we don't need ScreenRecording permission, let's remove this and the functions below

wispr 19.06.04/UI/Meeting/MeetingTranscriptView.swift
```diff
@@ -0,0 +1,235 @@
...
+            .disabled(meetingState.transcript.entries.isEmpty)
+        }
+        .padding(.horizontal, 16)
+        .padding(.vertical, 8)
+    }
+}
+
+// MARK: - SF Symbols Extensions
+
+private extension SFSymbols {
```
sebsto: - `waveform` already exists in the main `SFSymbols` enum (it's `menuBarProcessing`'s value and also `providerWhisper`)                                                     
  - `clipboard` duplicates `SFSymbols.copy` which is already "doc.on.doc"
  - These should be added to the central `SFSymbols.swift` file, not in a private extension

Sources/WisprApp/Services/MeetingAudioEngine.swift
```diff
@@ -0,0 +1,407 @@
...
+
+            let outputFrameCount = AVAudioFrameCount(Double(buffer.frameLength) * sampleRateRatio)
+            guard
+                let outputBuffer = AVAudioPCMBuffer(
+                    pcmFormat: targetFormat,
+                    frameCapacity: outputFrameCount
+                )
+            else { return }
+
+            nonisolated(unsafe) let inputBuffer = buffer
```
sebsto, line 209: This suppresses a Sendable warning for the `AVAudioPCMBuffer` captured into the converter closure. It's correct here (same synchronous callback), but worth a  one-line comment explaining why it's safe — this is exactly the kind of `nonisolated(unsafe)` that could be accidentally copied into an async context.

sebsto (changes requested):
Hey! Regarding the `exportTranscript()` method that uses `NSSavePanel` — it crashes because the app has no file-access entitlements, and `NSSavePanel` also has focus/activation issues in `.accessory` (menu-bar-only) apps.

The idiomatic SwiftUI replacement is `.fileExporter()` — it's a declarative view modifier that handles the save dialog automatically, requires **no entitlements** (user-selected locations are sandbox-safe by design), and avoids the window activation headaches.

Here's how it would work:

1. Create a simple `FileDocument` type:

```swift
import SwiftUI
import UniformTypeIdentifiers

struct TranscriptDocument: FileDocument {
    static let readableContentTypes: [UTType] = [.plainText]
    let text: String

    func fileWrapper(configuration: WriteConfiguration) throws -> FileWrapper {
        FileWrapper(regularFileWithContents: Data(text.utf8))
    }

    init(configuration: ReadConfiguration) throws {
        let data = configuration.file.regularFileContents ?? Data()
        text = String(decoding: data, as: UTF8.self)
    }

    init(text: String) { self.text = text }
}
```

2. Replace the `NSSavePanel` call with a `.fileExporter()` modifier on `MeetingTranscriptView`:

```swift
@State private var isExporting = false

// On the Export button:
Button("Export") { isExporting = true }

// On the view:
.fileExporter(
    isPresented: $isExporting,
    document: TranscriptDocument(text: meetingState.transcript.asPlainText()),
    contentType: .plainText,
    defaultFilename: "meeting-transcript"
) { result in
    if case .failure(let error) = result {
        Log.stateManager.error("Export failed: \(error.localizedDescription)")
    }
}
```

This removes the need for `exportTranscript()` on `MeetingStateManager` entirely — the view handles it declaratively. No entitlements, no focus issues, and it works on sandboxed builds too.

sebsto:
@gbrunoo I merged https://github.com/sebsto/wispr/pull/62 which changes the layout of the source files in the project. This will impact this PR, you'll need to move files around and use the new Xcode project layout.
There should be no conflict, just files moved to different directory.
Be sure to `pull main` and `merge main` before push any changes.

Sources/WisprApp/Services/MeetingAudioEngine.swift
gbrunoo: done as requested :)

sebsto:
@gbrunoo 

Bug: `.fileExporter()` still crashes — missing entitlement

**Steps to reproduce:**
1. Start a meeting transcription
2. Stop the meeting
3. Click the **Export** button (bottom-right of the transcript window)

**Crash message:**
> Unable to display save panel: your app has the User Selected File Read entitlement but it needs User Selected File Read/Write to display save panels. Please ensure that your app's target capabilities include the proper entitlements.

**Root cause:** The app currently has no entitlements file at all. SwiftUI's `.fileExporter()` still requires the `com.apple.security.files.user-selected.read-write` entitlement when the app uses hardened runtime (required for notarization).

**Fix:** Add an entitlements file to the wispr target with:

```xml
<key>com.apple.security.files.user-selected.read-write</key>
<true/>
```

Note: this app is **not** sandboxed (`com.apple.security.app-sandbox` is not set), but hardened runtime still enforces file entitlements for save panels. The entitlement just needs to be present in the code signing — no sandbox required.

Alternatively, if adding entitlements is too invasive for this PR, we could defer the Export feature and only ship Copy to Clipboard for the MVP (the Export button could be removed for now and added in a follow-up once entitlements are set up).

sebsto (changes requested):
Still crashing when exporting the meeting transcript

gbrunoo:
very strange... I have been unable to replicate the issue on my end.
I got a couple of friends testing the feature and they did not encounter the bug either

PR #67, opened 2026-06-22
Auto-save meeting transcripts to disk on stop

### LINKED ISSUE #66: Meeting transcript is lost when starting a new recording immediately after stopping
## Description

Meeting transcripts are held only in memory and are never persisted automatically. Starting a new meeting recording overwrites the previous transcript without saving it, causing irreversible data loss. If the user stops a recording and starts a new one (a common accidental action), the previous transcript is gone with no way to recover it.

## Environment

- **macOS**: 26.5.1 (25F80)
- **Wispr**: 1.10.0 (build 260528.280)
- **Hardware**: Mac15,6 (Apple Silicon)

## Steps to Reproduce

1. Open the meeting transcription window and start a recording.
2. Speak so that some transcript entries are generated.
3. Stop the recording.
4. Start a new recording immediately (without manually exporting or copying).
5. The previous transcript is permanently lost.

## Root Cause

In `MeetingStateManager.swift`:

- `startMeeting()` unconditionally resets the transcript, discarding the previous one without saving:

```swift
func startMeeting() async {
    guard meetingState == .idle else { return }
    ...
    transcript = MeetingTranscript()   // overwrites the previous transcript, no save
    errorMessage = nil
    ...
}
```

- `stopMeeting()` does not persist anything to disk — it only resets state and audio levels.

The transcript lives only in `MeetingStateManager.transcript` (in memory). The only ways to save are **manual**: the **Copy** button, or the SwiftUI `.fileExporter` in `MeetingTranscriptView`, which exports **plain text** (`asPlainText()`), not JSON. `MeetingTranscript` / `MeetingTranscriptEntry` are not `Codable`, and there is no auto-save or session history anywhere in the codebase.

## Suggested Fix

- Auto-save each completed session to disk on `stopMeeting()` (e.g. `~/Library/Application Support/wispr/Transcripts/` with a timestamped filename).
- Make `MeetingTranscript` / `MeetingTranscriptEntry` conform to `Codable` for clean serialization (JSON).
- (Optional follow-up) Add a session-history UI to browse and re-export past transcripts.

## Workaround

Before starting a new recording, manually preserve the current transcript using the **Copy** button or **Export** (plain text) action in the meeting window.

charlesrapp:
## Summary

Fixes #66 — meeting transcripts were held only in memory, so stopping a recording and immediately starting a new one discarded the previous session (the in-memory transcript is reset at the top of `startMeeting()`, and `stopMeeting()` never persisted anything).

This PR auto-saves every completed session to disk when a recording stops, so a prior transcript can no longer be lost by an accidental restart.

## Changes

- **`MeetingTranscript.swift`** — `MeetingSpeaker`, `MeetingTranscriptEntry`, and `MeetingTranscript` now conform to `Codable`.
- **`ModelPaths.swift`** — new `ModelPaths.transcripts` pointing at `<base>/transcripts/`, reusing the same sandbox-aware Application Support root as model storage.
- **`TranscriptStore.swift`** (new) — writes a completed transcript as pretty-printed JSON (ISO-8601 dates) to a timestamped file `meeting-yyyy-MM-dd_HH-mm-ss.json`. Empty transcripts are skipped.
- **`MeetingStateManager.stopMeeting()`** — persists the transcript via `TranscriptStore.save()` before returning to `.idle` (i.e. before any subsequent `startMeeting()` can reset it).

## Where transcripts are saved

`~/Library/Containers/com.stormacq.mac.wispr/Data/Library/Application Support/wispr/transcripts/` (sandboxed GUI), matching the existing model-storage convention in `ModelPaths`.

## Testing

- `swift build` succeeds.
- New `TranscriptStoreTests` (Swift Testing) — 4 tests, all passing:
  - Codable JSON round-trip preserves entries and plain-text rendering
  - `save` writes a decodable JSON file with the expected name shape
  - `save` skips an empty transcript
  - filename is derived from the transcript `startTime`
- Pre-existing `MeetingStateManagerTests` mic-permission failures are environment-dependent (they assume no microphone access) and reproduce on unmodified `main`; they are unrelated to this change.

## Notes / possible follow-ups

- This persists automatically and silently; a session-history UI to browse and re-export past transcripts would be a natural follow-up (mentioned in #66).
- Export from the UI remains plain text; on-disk persistence is JSON.

### DIFF
--- a/Sources/WisprApp/Models/MeetingTranscript.swift
+++ b/Sources/WisprApp/Models/MeetingTranscript.swift
@@ -8,13 +8,13 @@
 import Foundation
 
 /// Identifies the audio source / speaker in a meeting transcript.
-enum MeetingSpeaker: String, Sendable, Equatable, Hashable {
+enum MeetingSpeaker: String, Sendable, Equatable, Hashable, Codable {
     case you = "You"
     case others = "Others"
 }
 
 /// A single timestamped entry in a meeting transcript.
-struct MeetingTranscriptEntry: Identifiable, Sendable, Equatable {
+struct MeetingTranscriptEntry: Identifiable, Sendable, Equatable, Codable {
     let id: UUID
     let speaker: MeetingSpeaker
     let text: String
@@ -29,7 +29,7 @@ struct MeetingTranscriptEntry: Identifiable, Sendable, Equatable {
 }
 
 /// The full transcript of a meeting session.
-struct MeetingTranscript: Sendable, Equatable {
+struct MeetingTranscript: Sendable, Equatable, Codable {
     var entries: [MeetingTranscriptEntry] = []
     let startTime: Date
 
--- a/Sources/WisprApp/Services/MeetingStateManager.swift
+++ b/Sources/WisprApp/Services/MeetingStateManager.swift
@@ -127,6 +127,10 @@ final class MeetingStateManager {
 
         await meetingAudioEngine.stopCapture()
 
+        // Persist the completed session to disk before it can be overwritten
+        // by a subsequent startMeeting() (which resets `transcript`).
+        TranscriptStore.save(transcript)
+
         meetingState = .idle
         micLevel = 0
         systemLevel = 0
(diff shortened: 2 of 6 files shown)


sebsto (approved):
Thank you @charlesrapp for the PR

PR #71, opened 2026-07-16
perf(models): make model loading idempotent to avoid CoreML recompilation

### LINKED ISSUE #69: Model loading is not idempotent: redundant CoreML recompilation on repeated load/switch
### Problem

`WhisperService.loadModel(_:)` and `switchModel(to:)` (and the equivalent methods in `ParakeetService`) recreate the engine instance even when the requested model is already loaded and active. Recreating the `WhisperKit` instance re-runs CoreML/ANE AOT compilation and the `prewarm` pass, which is expensive and writes hundreds of MB to disk.

### Impact

- Unnecessary CPU and disk I/O on every redundant load/switch.
- On machines low on free disk space this compounds with macOS `cache_delete`, which purges the app container caches (where CoreML stores compiled artifacts) and can terminate the app.

### Steps to reproduce

1. Load a model.
2. Call `loadModel` (or `switchModel`) again with the same model id.
3. Observe a second `WhisperKit` initialization and prewarm instead of a no-op.

### Proposed fix

Make `loadModel`/`switchModel` idempotent: skip the work when the requested model is already active with a live instance. Add a short-circuit at `CompositeTranscriptionEngine` as well.

I have a change ready and will open a PR.

charlesrapp:
Closes #69

  loadModel/switchModel recreated the engine instance even when the requested model was already active, re-running CoreML/ANE compilation and the prewarm pass (hundreds of MB written to disk each time).

  This makes loading idempotent in WhisperService and ParakeetService, and short-circuits CompositeTranscriptionEngine.loadModel when the model is already active on the right engine. Adds unit tests asserting the underlying engine is not re-invoked for the already-active model.

### DIFF
--- a/Sources/WisprCore/Services/CompositeTranscriptionEngine.swift
+++ b/Sources/WisprCore/Services/CompositeTranscriptionEngine.swift
@@ -96,6 +96,11 @@ public actor CompositeTranscriptionEngine: TranscriptionEngine {
             throw WisprError.modelLoadFailed("No engine found for model \(modelName)")
         }
 
+        // Already loaded on the right engine — skip the unload/reload cycle.
+        if activeEngineIndex == idx, await engines[idx].activeModel() == modelName {
+            return
+        }
+
         // Unload the previous engine to free memory before loading the new one.
         if let currentIdx = activeEngineIndex, currentIdx != idx {
             await engines[currentIdx].unloadCurrentModel()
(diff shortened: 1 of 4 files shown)


sebsto:
Thank you Charles for having catch this.
I made a tiny change to use the Swift idiomatic `guard` statement instead of `if`.

PR #78, opened 2026-07-20
Add custom vocabulary correction for proper nouns and tool names

omansour:
## Problem

ASR models mangle proper nouns they've never seen: "kubectl" comes out as
"cube cuttle", "PyTorch" loses its casing, rare names get split into several
words. There's no way to teach Wispr the spelling of words that matter to you.

## Solution

A **Custom Vocabulary** list (Settings → After Transcription): any run of 1–3
transcribed words that *sounds like* a vocabulary entry is rewritten to its
canonical spelling. Matching is phonetic (per-language key + edit-distance
threshold, user's language + English so both pronunciations of "PyTorch" work)
and purely lexical — engine-agnostic, applied to both dictation and meeting
transcripts. Details in `VocabularyCorrector.swift`.

## Try it

Enable **Custom Vocabulary** in Settings, add `kubectl` and `PyTorch`, then
dictate *"I deployed the pytorch model with kubectl"* — both come out correctly
spelled, neighbouring words untouched.

Covered by unit tests (corrector, settings persistence, pipeline wiring).

**Note:** first commit is an unrelated one-line fix — `TranscriptStoreTests` no
longer compiled after the `.others(speakerIndex:)` enum change. Can split it
out if preferred.

### DIFF
--- a/Sources/WisprApp/Services/StateManager.swift
+++ b/Sources/WisprApp/Services/StateManager.swift
@@ -252,6 +252,21 @@ final class StateManager {
         return FillerWordCleaner.clean(text)
     }
 
+    // MARK: - Custom Vocabulary Correction
+
+    /// Corrects mis-transcribed proper nouns against the user's custom vocabulary
+    /// when the setting is enabled and the list is non-empty.
+    func applyVocabularyCorrection(to text: String) -> String {
+        guard settingsStore.customVocabularyEnabled,
+              !settingsStore.customVocabulary.isEmpty,
+              !text.isEmpty else { return text }
+        return VocabularyCorrector.correct(
+            text,
+            vocabulary: settingsStore.customVocabulary,
+            languageCode: settingsStore.languageMode.languageCode
+        )
+    }
+
     // MARK: - AI Text Correction
 
     /// Applies AI text correction if enabled and available.
@@ -321,7 +336,11 @@ final class StateManager {
         // AI text correction step (after filler removal, before suffix)
         let corrected = await applyAITextCorrection(to: cleaned)
 
-        let finalText = applyAutoSuffix(to: corrected)
+        // Custom vocabulary correction — runs after AI correction so it has the
+        // final say on the spelling of known proper nouns.
+        let vocabCorrected = applyVocabularyCorrection(to: corrected)
+
+        let finalText = applyAutoSuffix(to: vocabCorrected)
 
         do {
             try await textInsertionService.insertText(finalText)
--- a/Sources/WisprApp/Services/StateManager.swift
+++ b/Sources/WisprApp/Services/StateManager.swift
@@ -230,7 +230,8 @@ final class StateManager {
 
                 self.soundFeedback.play(.recordingStopped)
 
-                await self.insertTranscribedText(finalResult.text)
+                await self.insertTranscribedText(
+                    finalResult.text, detectedLanguage: finalResult.detectedLanguage)
             } catch {
                 guard !Task.isCancelled else { return }
                 Log.stateManager.warning("EOU monitoring failed: \(error.localizedDescription)")
@@ -256,14 +257,18 @@ final class StateManager {
 
     /// Corrects mis-transcribed proper nouns against the user's custom vocabulary
     /// when the setting is enabled and the list is non-empty.
-    func applyVocabularyCorrection(to text: String) -> String {
+    ///
+    /// The language used for phonetic matching is the configured one when pinned
+    /// or specific; in auto-detect mode (where `languageMode.languageCode` is nil)
+    /// it falls back to the language the engine detected for this transcription.
+    func applyVocabularyCorrection(to text: String, detectedLanguage: String? = nil) -> String {
         guard settingsStore.customVocabularyEnabled,
               !settingsStore.customVocabulary.isEmpty,
               !text.isEmpty else { return text }
         return VocabularyCorrector.correct(
             text,
             vocabulary: settingsStore.customVocabulary,
-            languageCode: settingsStore.languageMode.languageCode
+            languageCode: settingsStore.languageMode.languageCode ?? detectedLanguage
         )
     }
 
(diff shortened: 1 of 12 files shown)


Sources/WisprApp/Services/StateManager.swift
```diff
@@ -252,6 +253,25 @@ final class StateManager {
...
+    /// or specific; in auto-detect mode (where `languageMode.languageCode` is nil)
+    /// it falls back to the language the engine detected for this transcription.
+    func applyVocabularyCorrection(to text: String, detectedLanguage: String? = nil) -> String {
+        guard settingsStore.customVocabularyEnabled,
+              !settingsStore.customVocabulary.isEmpty,
+              !text.isEmpty else { return text }
+        return VocabularyCorrector.correct(
+            text,
+            vocabulary: settingsStore.customVocabulary,
+            languageCode: settingsStore.languageMode.languageCode ?? detectedLanguage
```
sebsto, line 271: **Auto-detect fallback is a no-op for Parakeet.** `languageMode.languageCode ?? detectedLanguage` only recovers a language when `detectedLanguage` is non-nil, but `ParakeetService` always returns `detectedLanguage: nil` (see `ParakeetService.swift:160/193/498`) — only `WhisperService` populates it. So for a Parakeet model in auto-detect mode, `languageCode` is `nil` and `VocabularyCorrector` folds with English rules only.

This affects every Parakeet-driven path — the EOU dictation path (line 234) and the meeting mic/system paths (`MeetingStateManager.swift:232`/`263`) — while only the batch Whisper path (line 538) actually benefits. Since the commit message and the doc comment above say it "falls back to the language the engine detected," it'd be worth either making Parakeet report a detected language or softening the claim so the behavior isn't surprising.

Sources/WisprApp/Utilities/VocabularyCorrector.swift
```diff
@@ -0,0 +1,422 @@
...
+            ("ck", "k"),
+            ("ch", "k"),
+            ("ou", "u"),
+            ("oo", "u"),
+            ("ee", "i"),
+            ("ea", "i"),
+            ("ai", "e"),
+            ("ei", "e"),
+            ("au", "o"),
+            ("eau", "o"),
```
sebsto, line 291: **This `("eau", "o")` rule is unreachable.** The digraph list is applied in order, and both `("ea", "i")` and `("au", "o")` run earlier — so by the time this entry is reached, `"eau"` has already been rewritten (`"eau"` → `"iu"` via the `"ea"` rule). The intended "eau ≈ /o/" normalization never fires for any input. If it matters for names like "Bordeaux", move this rule ahead of `"ea"`/`"au"`; otherwise it can be dropped as dead code.

Sources/WisprApp/Utilities/VocabularyCorrector.swift
```diff
@@ -0,0 +1,422 @@
...
+
+        var best: (canonical: String, score: Double)?
+        for entry in entries {
+            var entryScore = 0.0
+            for language in languages {
+                guard let wk = windowKeys[language], let ek = entry.keysByLanguage[language] else { continue }
+                // Keys too short for the threshold to discriminate: only accept a
+                // case-insensitive spelling match (still fixes casing, e.g. "go" → "Go").
+                if ek.count < minKeyLengthForFuzzyMatch || wk.count < minKeyLengthForFuzzyMatch {
+                    if joined.caseInsensitiveCompare(entry.canonical) == .orderedSame {
```
sebsto, line 198: Minor: the short-key fallback compares `joined` (alphanumeric cores concatenated, spaces stripped) against `entry.canonical` (which may contain spaces). So a multi-word vocabulary entry whose phonetic key is under `minKeyLengthForFuzzyMatch` can never match even on an exact spelling — e.g. `["A B"]` vs input `"a b"`: `"ab".caseInsensitiveCompare("A B")` is not `.orderedSame`. Comparing against a space-stripped form of the canonical would close the gap. Edge case, non-blocking.

Sources/WisprApp/UI/Components/VocabularyEditorView.swift
```diff
@@ -0,0 +1,79 @@
...
+                .accessibilityLabel("Add word")
+            }
+
+            // Existing words.
+            if vocabulary.isEmpty {
+                Text("No words yet. Add client or product names that are often mis-transcribed.")
+                    .font(.caption)
+                    .foregroundStyle(.secondary)
+            } else {
+                ForEach(Array(vocabulary.enumerated()), id: \.offset) { index, word in
```
sebsto, line 40: Nit: `id: \.offset` keys each row by its position, so on a mid-list removal SwiftUI tracks identity by index rather than content — the delete animation targets the wrong row and any future per-row `@State` would carry over incorrectly. The underlying data stays correct thanks to the bounds-checked `remove(at:)`, so this is cosmetic. Since `addWord` already guarantees entries are unique case-insensitively, `id: \.self` would be both correct and simpler here.

sebsto (commented):
Thanks so much for this — and welcome! 🎉 This is a really nicely put together first contribution to Wispr: a genuinely useful feature, clean separation between the pure `VocabularyCorrector` utility and the wiring, thoughtful multilingual/phonetic handling, and solid test coverage (the tie-breaking and short-word cases in particular show real care). I especially appreciate that you addressed the earlier review points (auto-detect language, short words, meeting pipeline) in a follow-up commit.

I reviewed the whole diff and the core algorithm holds up well — tokenizer/separator alignment, greedy non-overlap selection, the `core`/`affixes` index arithmetic, and `SettingsStore` persistence symmetry all check out. I left a few inline comments; none are blockers. The main one worth a look is the auto-detect language fallback: it's effectively a no-op for the Parakeet engine (which never reports a detected language), so every streaming/EOU/meeting path currently falls back to English-only matching rather than the detected language the commit message describes. The others are minor (a dead phonetic rule, a narrow multi-word edge case, and a SwiftUI list-identity nit).

Thanks again for the effort and your interest in the project — happy to help get this over the line.

sebsto (changes requested):
some minor feedback

hCoureau:
Hi @omansour and @sebsto — this is a valuable feature and the review notes look very actionable. Since the branch is now conflicting with main, would you be comfortable with me reviving it in a separate follow-up PR based on current main? I would retain the original attribution, link the work back to this PR, and address the outstanding review feedback rather than replacing the contribution.

sebsto:
Thank you @hCoureau 
Yes please, go ahead as the original contributor never replied to the code review. Let’s proceed

sebsto:
Superseded by #102

PR #85, opened 2026-07-24
feat(meeting): reveal saved transcripts folder from Settings

### LINKED ISSUE #82: Make saved meeting transcripts discoverable from Settings
## Context

Since #67, meeting transcripts are auto-saved on Stop by `TranscriptStore.save(_:)` as JSON under the app's sandbox container:

```
~/Library/Containers/com.stormacq.mac.wispr/Data/Library/Application Support/wispr/transcripts/meeting-<start>.json
```

That location is great for durability (guaranteed writable under the sandbox, never lost), but users can't realistically find it: `~/Library` is hidden, the path is buried inside the container, and the files are JSON (not human-readable).

## Proposal

Surface the transcripts from **Settings** rather than adding a fully custom save location:

1. Add an **"Open Transcripts Folder"** button in the meeting section of `SettingsView`, using `NSWorkspace.shared.open(TranscriptStore.directory)`. The app opens its **own** container folder, so no extra entitlement is required.
2. Show the folder path as small, selectable secondary text for reference.
3. (Optional) Also write a human-readable `.md` (or `.txt`) next to the JSON on each save, so opening the folder is actually useful. JSON stays the source of truth for reload/robustness.

## Sandbox note (why not a free-form custom path)

The app is sandboxed with `user-selected.read-write` only. A user-chosen destination stored as a plain path string would be denied at write time. Doing it properly needs a security-scoped bookmark (one-time folder picker, persisted, resolved on each save). That is heavier, so I'd suggest deferring it unless there's demand (iCloud Drive / Obsidian / etc.).

## Questions

- Button + displayed path only, or also write a Markdown/text copy alongside the JSON?
- Is a configurable folder (security-scoped bookmark) something you want now, or is opening the default folder enough for a first step?

Happy to open a PR once you've picked a direction.

charlesrapp:
Closes #82

Meeting transcripts are auto-saved on Stop (#67) as JSON inside the app's sandbox container, which users can't realistically find (`~/Library` is hidden and the path is buried in the container).

This adds discoverability from **Settings → Meeting**, per your call on #82 (button + displayed path only, no custom location, sandbox left intact):

- An **"Open Folder"** button that reveals the transcripts directory in Finder. It opens the app's **own** container folder via `NSWorkspace`, so no extra entitlement is needed, and creates the folder first if no meeting has been saved yet.
- The home-abbreviated path shown below it as small, selectable text.

Deliberately **not** included: a configurable save location (would require a security-scoped bookmark) and a Markdown/text copy — keeping the change minimal and the sandbox untouched.

### Testing

- `xcodebuild build` and `WisprTests` pass (the only failures are the pre-existing `TextCorrectionTokenLeakTests`, which need Apple Intelligence at runtime and are unrelated).

### DIFF
--- a/Sources/WisprApp/UI/Settings/SettingsView.swift
+++ b/Sources/WisprApp/UI/Settings/SettingsView.swift
@@ -6,7 +6,6 @@
 //  Recognition, After Transcription, Feedback, and General.
 //
 
-import AppKit
 import SwiftUI
 import WisprCore
 import os
@@ -89,6 +88,7 @@ struct SettingsView: View {
     @Environment(HotkeyMonitor.self) private var hotkeyMonitor: HotkeyMonitor
     @Environment(TextCorrectionService.self) private var textCorrectionService:
         TextCorrectionService
+    @Environment(\.openURL) private var openURL
 
     @State private var audioDevices: [AudioInputDevice] = []
     @State private var whisperModels: [ModelInfo] = []
@@ -437,8 +437,10 @@ struct SettingsView: View {
 
     /// Home-abbreviated path to the folder where meeting transcripts are saved.
     private var transcriptsFolderPath: String {
-        (TranscriptStore.directory.path(percentEncoded: false) as NSString)
-            .abbreviatingWithTildeInPath
+        let path = TranscriptStore.directory.path(percentEncoded: false)
+        let home = FileManager.default.homeDirectoryForCurrentUser.path(percentEncoded: false)
+        guard path.hasPrefix(home) else { return path }
+        return "~" + path.dropFirst(home.count)
     }
 
     /// Reveals the transcripts folder in Finder, creating it first if no meeting
@@ -446,7 +448,7 @@ struct SettingsView: View {
     private func openTranscriptsFolder() {
         let dir = TranscriptStore.directory
         try? FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
-        NSWorkspace.shared.open(dir)
+        openURL(dir)
     }
 
     // MARK: - General Section
--- a/Sources/WisprApp/UI/Settings/SettingsView.swift
+++ b/Sources/WisprApp/UI/Settings/SettingsView.swift
@@ -6,6 +6,7 @@
 //  Recognition, After Transcription, Feedback, and General.
 //
 
+import AppKit
 import SwiftUI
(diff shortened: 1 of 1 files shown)


sebsto:
Thank you @charlesrapp for this PR.
I would like to a void using Appkit and NXxxxx API as much as possible. This project leans forward a pure SwiftUI-style implementation.

Maybe it's a rule I should add in CLAUDE.md.

Sources/WisprApp/UI/Settings/SettingsView.swift
```diff
@@ -6,6 +6,7 @@
 //  Recognition, After Transcription, Feedback, and General.
 //
 
+import AppKit
```
sebsto: The code must avoid using NSxxxx API. This PR introduces two (NSString and NSWorkspace), In Swift UI, I propose to use something like 

```swift
// No `import AppKit` needed — just SwiftUI + Foundation

// MARK: - Transcripts Folder

/// Home-abbreviated path to the folder where meeting transcripts are saved.
private var transcriptsFolderPath: String {
    let path = TranscriptStore.directory.path(percentEncoded: false)
    let home = FileManager.default.homeDirectoryForCurrentUser.path(percentEncoded: false)
    if path.hasPrefix(home) {
        return "~" + path.dropFirst(home.count)
    }
    return path
}

/// Reveals the transcripts folder in Finder.
@Environment(\.openURL) private var openURL

private func openTranscriptsFolder() {
    let dir = TranscriptStore.directory
    try? FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
    openURL(dir)
}`

sebsto (changes requested):

charlesrapp:
Thanks for the review — good call, and agreed on keeping the codebase SwiftUI-first.

Fixed in 6094969, following your suggestion:

- `NSWorkspace.shared.open` → `@Environment(\.openURL)`
- `NSString.abbreviatingWithTildeInPath` → plain Foundation prefix check against `FileManager.default.homeDirectoryForCurrentUser`
- `import AppKit` removed from `SettingsView.swift`

`SettingsView.swift` now has no `NS*` symbol left. Build passes.

Happy to open a separate PR adding the rule to `CLAUDE.md` if you want it documented.

sebsto:
Make sure to resend from main !

PR #87, opened 2026-07-28
Make a backgrounded meeting visible, stoppable, and never lost on quit

el-pedrito:
Closing the meeting window does not stop the meeting — `windowWillClose` only clears the visibility flags, and recording continues. That is the right behaviour for a menu bar app, but nothing else in the app acknowledged it, which left three problems.

### The transcript was lost on quit

`applicationWillTerminate` calls `cancelRecording()`, which cancels the task group and saves nothing. `TranscriptStore.save` ran only from `stopMeeting()`. So: start a meeting, close the window, keep talking, quit from the menu — the entire session is destroyed with no warning.

`finalizeForTermination()` now writes the transcript before cancelling. It is synchronous because `stopMeeting()` is `async` and the app is torn down before the suspension would resume. Saving is silent, and empty transcripts are still skipped.

### A running meeting was invisible

`startObservingState()` tracked `stateManager.appState` but not `meetingState`, so a recording meeting with the window closed produced no signal at all: the menu title was unchanged, and there was no way to stop it without reopening the window.

The menu item now shows the elapsed time with a filled icon, and a "Stop Meeting" item appears while recording.

### Reopening from the menu could be a dead click

`show()` began with `guard !isVisible else { return }`. Clicking the menu item while the window was open but buried behind a fullscreen app therefore did nothing. It now brings the panel to the front, matching the already-visible handling in `openSettings()`.

### Two window annoyances

`positionPanel` ran on every show, snapping the window back to the bottom-right corner and discarding the user's own placement. It now runs only on first show, with `setFrameAutosaveName` persisting position and size across launches.

`minSize` was 320x300 while the content declares `.frame(minWidth: 360, minHeight: 400)`, so the window could be dragged smaller than its content and clip the transcript.

### Testing

`swift build` and the full `swift test` suite pass (536 tests). `xcodebuild` build passes. Two new tests cover `finalizeForTermination`: the no-op when idle, and the persistence path, which compares the transcripts directory before and after to assert a file is actually written, then cleans up after itself.

Verified manually: close the window mid-meeting, speak, quit from the menu, and the JSON is on disk with the speech that arrived after the window was closed.

### Note

No behaviour changes when the window is open and the app is quit normally through `stopMeeting()` — that path already saved.

### DIFF
--- a/Sources/WisprApp/Services/MeetingStateManager.swift
+++ b/Sources/WisprApp/Services/MeetingStateManager.swift
@@ -70,6 +70,13 @@ final class MeetingStateManager {
     /// Whether diarization is active for the current session (set at startMeeting).
     private var diarizationActive = false
 
+    /// Guards against a second save while `stopMeeting()` is suspended.
+    ///
+    /// `stopMeeting()` awaits before it flips `meetingState` to `.idle`, so a quit
+    /// during that window would let `finalizeForTermination()` see a still-recording
+    /// session and write a second file.
+    private var isStopping = false
+
     /// Guards against re-entrant `startMeeting()` while capture is being set up
     /// (the window between the first `await` and `meetingState = .recording`).
     private var isStarting = false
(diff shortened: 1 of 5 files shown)


el-pedrito:
Superseded by #88, which combines this window-lifecycle work with the transcript-history feature that builds on it into a single PR (kept as separate atomic commits). Closing this to avoid a duplicate/overlapping diff.

PR #88, opened 2026-07-28
Meeting mode: keep the window in the background, then browse and name past transcripts

el-pedrito:
Two related improvements to meeting mode, kept as separate atomic commits so they can be reviewed one at a time:

1. Make a backgrounded meeting visible, stoppable, and never lost on quit
2. Browse, reopen, name, and delete past meeting transcripts (with window ergonomics)

The second builds on the first, so they are proposed together here.

---

## Part 1 — Backgrounded meetings: visible, stoppable, never lost

Closing the meeting window does not stop the meeting — `windowWillClose` only clears the visibility flags, and recording continues. That is the right behaviour for a menu bar app, but nothing else in the app acknowledged it, which left three problems.

### The transcript was lost on quit

`applicationWillTerminate` calls `cancelRecording()`, which cancels the task group and saves nothing. `TranscriptStore.save` ran only from `stopMeeting()`. So: start a meeting, close the window, keep talking, quit from the menu — the entire session is destroyed with no warning.

`finalizeForTermination()` now writes the transcript before cancelling. It is synchronous because `stopMeeting()` is `async` and the app is torn down before the suspension would resume. Saving is silent, and empty transcripts are still skipped.

### A running meeting was invisible

`startObservingState()` tracked `stateManager.appState` but not `meetingState`, so a recording meeting with the window closed produced no signal at all: the menu title was unchanged, and there was no way to stop it without reopening the window.

The menu item now shows the elapsed time with a filled icon, and a "Stop Meeting" item appears while recording.

### Reopening from the menu could be a dead click

`show()` began with `guard !isVisible else { return }`. Clicking the menu item while the window was open but buried behind a fullscreen app therefore did nothing. It now brings the panel to the front, matching the already-visible handling in `openSettings()`.

### Two window annoyances

`positionPanel` ran on every show, snapping the window back to the bottom-right corner and discarding the user's own placement. It now runs only on first show, with `setFrameAutosaveName` persisting position and size across launches.

`minSize` was 320x300 while the content declares `.frame(minWidth: 360, minHeight: 400)`, so the window could be dragged smaller than its content and clip the transcript.

### Note

No behaviour changes when the window is open and the app is quit normally through `stopMeeting()` — that path already saved.

---

## Part 2 — Browse, reopen, name, and delete past transcripts

`TranscriptStore` saved every session as JSON but exposed only `save()`. Sessions were written to disk and then invisible to the user: no way to list, reopen, rename, or delete them. #82 raised the same problem and #85 shipped a button to reveal the folder; this goes the rest of the way inside the app.

### What this adds

- `TranscriptStore.list()`, `load(_:)`, `delete(_:)`, `save(_:to:)`
- A history sidebar in the **existing** meeting window, with the live session pinned on top. A separate window would split the user's mental model in two.
- Sessions can be named ("Sprint review" rather than a date), renamed, and reverted to their date
- Speakers can be named, which is most useful after the meeting — which is when you actually know who was who
- Copy, Export, Show in Finder, Delete

### Window ergonomics

Three refinements so the window reads clearly whether a meeting is live, idle, or being reviewed:

- **Start/Stop lives in the toolbar**, next to the sidebar-collapse button, as a prominent tinted button (accent to start, red to stop). It used to sit inside the transcript header, which meant it vanished the moment you selected an archived transcript — you had to go back to the live view just to start or stop. It is now reachable from anywhere, at the very top.
- **Stopping hands off cleanly.** `stopMeeting()` clears the live transcript and the sidebar jumps to the freshly saved session. Previously the finished meeting stayed rendered under "Current Session" *and* appeared as a new history row — the same transcript in two places, with no cue that recording had ended. "Current Session" now returns to an empty slate, and the archived session is selected automatically.
- **Rename fields focus on open.** Both the session-title field (sidebar) and the speaker-name field (roster) grab keyboard focus as they appear, so you can type immediately instead of clicking into the box first.

The transcript header is hidden while the live session is idle, since with Start/Stop moved out it would otherwise be an empty bar.

### Screenshots

- `history-window-overview.png` — sidebar, a reopened session, speaker chips, and the toolbar Start button
- `history-current-session-empty.png` — the clean-slate "Current Session" after a meeting ends
- `history-context-menu-rename.png` — rename / reveal / delete actions
- `history-rename-inline-focus.png` — inline rename field, focused on open

### Speaker names — please read this before testing

`speakerNames` maps the 0-based diarization index to a name, resolved at display time and never stored per entry, so renaming relabels all of a speaker's existing entries and every future one — no entry rewritten, no SwiftUI identity churn.

**The speaker roster is empty unless `meetingDiarizationEnabled` is on**, which is off by default. Without diarization a transcript contains only "You" and "Others", so there are no individual speakers to name and no chips are shown. This is expected, not a bug — I have deliberately not changed the shipped default, since that is a maintainer's call. Turn the setting on to exercise this part.

Naming lives on a roster above the transcript rather than in a per-row context menu: the rows auto-scroll during a live meeting, which makes a 56pt badge a moving target, and a name is a property of the roster, not of one line.

### Backward compatibility

`MeetingTranscript` gains an explicit `init(from:)`. This is not optional politeness: the compiler-synthesized initializer calls `decode(_:forKey:)` for every non-optional stored property and **ignores the default-value expression**, so relying on `= [:]` would have made every transcript already on disk fail to decode with `keyNotFound`. Two tests decode real legacy files — one without `speakerNames`, one without `title` — to lock this down.

### Live and archived transcripts are kept separate

The view previously read `meetingState.transcript` directly, and `startMeeting()` assigns a fresh transcript. Starting a meeting while reading a past one would therefore have wiped what was on screen. `MeetingHistoryStore` owns the loaded transcript and the view renders whichever is selected; auto-scroll follows the live session only. A test covers exactly this scenario.

### Off the main actor

`TranscriptStore` becomes `nonisolated`, as do the three model types. Under this package's `.defaultIsolation(MainActor.self)` an unannotated conformance is a *MainActor-isolated conformance*, so decoding could not have compiled in a nonisolated context. Without it, listing a hundred transcripts would decode them all on the UI thread. `stopMeeting()`'s save also moves off the main thread while still being awaited, which preserves ordering against a subsequent `startMeeting()`.

### Smaller fixes this needed

- `duration` is now-relative, so a transcript reopened a week later reported the time elapsed since it started rather than the meeting's length. Added `contentDuration`, derived from the entries.
- Two meetings starting in the same second produced the same filename and the second silently overwrote the first. A uniquing suffix is added.
- A corrupt file now appears as an unreadable, deletable row instead of being skipped — a silent disappearance looks like data loss.
- Deletion is always confirmed, and says why: wispr keeps no audio to re-transcribe from.

### Sizing

The window grows to 760x560 (min 620x420) to fit the sidebar. The transcript rows spend roughly 154pt on the timestamp and speaker columns before any text, so a narrower window leaves nothing readable beside the sidebar.

---

## Testing

Full `swift test` suite passes, `xcodebuild` Release build passes. New files are registered in `project.pbxproj`, which lists sources explicitly rather than globbing as SwiftPM does — without that, an Xcode build fails with `cannot find 'MeetingHistoryStore' in scope` even though `swift build` succeeds.

Two tests cover `finalizeForTermination` (the no-op when idle, and the persistence path that asserts a file is written then cleans up). The meeting and history suites cover the store operations, the live/archived separation, and the stop→save path. The window-ergonomics changes are UI-level and were verified manually on a signed Release build: starting/stopping from the toolbar, the clean-slate handoff on stop, and focus landing in both rename fields on open.

Verified manually: close the window mid-meeting, speak, quit from the menu, and the JSON is on disk with the speech that arrived after the window was closed.

## Out of scope

- Full-text search across transcripts
- Retention / auto-purge policy — deleting user data at launch felt like the wrong default to introduce unasked
- Audio retention


el-pedrito:
history-window-overview
history-current-session-empty
history-context-menu-rename
history-current-session-empty

el-pedrito:
## Update — transcript storage and history management

Nine further commits on top of Part 2. Same theme, so they are here rather than in a
separate PR, but they are a distinct third chunk and can be reviewed as one.

**Choosing where transcripts are stored.** They were pinned to the app container, which
is awkward to reach and impossible to put somewhere synced. Settings gains `Change…` and
`Use Default`. The app is sandboxed, so the choice is persisted as a security-scoped
bookmark rather than a path — a path resolves after relaunch to a folder the app may no
longer open. `ENABLE_USER_SELECTED_FILES` was already `readwrite`, so no entitlement
change. A stale bookmark is refreshed; an unusable one is cleared and the default
restored with the reason shown, since writing transcripts somewhere unreachable is worse
than writing them to the default. Files already saved elsewhere are left where they are.

**Deleting several at once.** Sidebar selection becomes a `Set`, while display still
follows a single row — there is no sensible "current transcript" for a multi-selection.
Right-clicking inside a selection acts on all of it, outside it acts on that row alone,
as Finder does. Batch deletion continues past an individual failure and reports them
together, so it is never unclear which transcripts actually went.

**Renaming a session renames its file.** A name lived only inside the JSON, so a folder
of `meeting-<timestamp>.json` gave no clue what anything was. The `meeting-` prefix and
timestamp deliberately stay in front: `list()` filters on that prefix, so a free-form
name would make the transcript vanish from the sidebar. Since summaries are keyed by URL,
the rename also remaps the selection, or the transcript being read would look deleted.

**Following the folder on disk.** Deleting a transcript in Finder left a row that opened
nothing. A `DispatchSource` watch drives a rescan, debounced because one Finder action
emits several vnode events. Scoped to the meeting window — opening it refreshes anyway,
so the list is never stale on screen, but this is not a permanent watch.

The reverse rename direction (Finder → app) is implemented and then reverted in
`d1cd4c7`. It could not be made to hold its shape: the slug is lossy so punctuation did
not survive a round trip, adopting the filename meant writing to files during a
read-only scan, and it only ran while the window was open. Left in history rather than
squashed, in case anyone wants to try again.

### Touching things outside this feature — please object if unwanted

- `SettingsView.swift`: removed the `Divider()` from 02a14b7. Inside a `Form` on macOS it
  takes its own row and renders as a short vertical stub, and it was the only one in the
  view — every other section uses the `Form`'s own separators.
- `MeetingStateManagerTests.swift`: `testFinalizeForTerminationSavesInProgress` found its
  file by diffing the transcripts directory. Several suites share that one real directory
  and run in parallel, so it both mis-counted and deleted other suites' files. It now
  locates its own transcript by content — start time cannot be the key, ISO-8601 encoding
  drops the sub-second part on the way to disk.
- `project.pbxproj`: registers the new sources, without which the Xcode build fails while
  `swift build` passes. `CURRENT_PROJECT_VERSION` deliberately untouched — bumping it is a
  release action.
- `Makefile` / `.gitignore`: additions only. `make run` builds Release, unregisters any
  other bundle claiming the identifier, restarts, and reports what is running. This exists
  because a stale build is invisible: an incremental build leaves the bundle's mtime
  untouched and the advertised build number only moves on `make archive`, so a Debug
  bundle registered beside the Release one gets launched by Spotlight while looking
  identical. Signing is detected, not forced — ad-hoc only when the keychain holds no
  identity, so your signing is never overridden. These introduce `.xcbuild/` as the
  derived-data path, which is not a convention here; happy to drop them.

### Test isolation this required

The transcript suites share one real directory and run in parallel, while the new
location tests redirect that directory process-wide — so a redirect landed inside another
suite's test, whose `refresh()` then scanned the wrong folder and reported the file it had
just written as missing. A different test failed on almost every run. A `TestScoping` trait
now gives those suites exclusive use of the directory, and test start times come from a
shared counter set in the year 2000 so two tests cannot derive the same filename, nor
collide with a real recording.

### Testing

31 new tests: slug and filename rules, bookmark round-trip and the fallbacks when a saved
folder is unusable, folder-change broadcast, batch deletion including partial failure, and
the watcher against real external create / delete / rename. `swift test` and an
`xcodebuild` Release build both pass.

Two caveats rather than a clean claim:

- `AudioEngine audio level stream terminates on stop` fails in a full run, passes in
  isolation. Pre-existing — the full suite fails identically at unmodified `main` in a
  clean worktree.
- Once in about ten full runs, `A renamed file is still found by list()` failed. It did not
  reproduce in six consecutive runs afterwards nor in isolation, so I cannot say whether it
  is a residual effect of the contention described above or something else.

Verified by hand: picking a folder and seeing new transcripts land there, and renaming a
session renaming the file in Finder. Not verified by hand, no signing certificate on this
machine so no UI automation: ⌘/⇧-click range selection, and the Finder-deletion round trip.

gbrunoo:
Hi Pierre, nice work on this, the history sidebar and the "never lost on quit" fix are both very welcome. I built the branch locally and spent some time in the meeting window. Three things on the toolbar, all in the same file, and I have a verified diff for all of them at the bottom.

## 1. The sidebar toggle moves around (needs fixing)

With the sidebar open, the collapse icon sits just right of "Start Meeting". Collapse the sidebar and it jumps to the far right of the title bar. It should stay in one place in both states.

Cause: `MeetingTranscriptView` uses `NavigationSplitView` without a `columnVisibility` binding, so SwiftUI supplies the collapse button itself. That system button sits in the leading group next to your `.navigation` items while the sidebar is open, then relocates to the trailing edge when it collapses. Since the button isn't yours, its placement isn't yours to control either.

Fixing it takes three things together:

1. Bind `columnVisibility`, so a button of your own can drive the collapsed state.
2. Remove the system toggle with `.toolbar(removing: .sidebarToggle)`, applied **to the sidebar column's content**, not to the `NavigationSplitView`. Worth calling out: at the outer level it compiles happily and does nothing, and you end up with two toggle icons side by side. That one cost me a build to work out.
3. Add your own toggle as a `ToolbarItem(placement: .navigation)` ahead of Start Meeting.

## 2. The Start Meeting restyle looks unintentional

`68950f8` moved Start/Stop into the toolbar, which also swapped it from a `.plain` button with a soft tinted rounded rect to a `.borderedProminent` pill.

The move itself is right: the old button lived in the live session's header and vanished as soon as an archived transcript was selected, so you had to detour back to "Current Session" just to start or stop a meeting. Toolbar placement fixes that properly.

The new look seems more like what `.borderedProminent` gives you by default than a deliberate choice, and it makes Start Meeting the loudest thing in the window. The good news is the original design does still work in a toolbar: the diff below restores your header styling verbatim (lifted from `68950f8^`), just hosted in the toolbar item. Happy to defer to you if you prefer the pill.

## 3. Add a gap between the toggle and Start/Stop

Once both sit in the leading group they render flush against each other and read as one segmented control. 8pt of leading padding on Start/Stop separates them.

One dead end worth saving you the trouble: `ToolbarSpacer(.fixed, placement: .navigation)` is the obvious tool here and it compiles fine on the macOS 26.2 target, but it produced no gap at all. I measured the button coordinates before and after and they were identical. Plain padding is what actually worked.

## The diff

```diff
diff --git a/Sources/WisprApp/UI/Meeting/MeetingTranscriptView.swift b/Sources/WisprApp/UI/Meeting/MeetingTranscriptView.swift
--- a/Sources/WisprApp/UI/Meeting/MeetingTranscriptView.swift
+++ b/Sources/WisprApp/UI/Meeting/MeetingTranscriptView.swift
@@ -31,6 +31,11 @@ struct MeetingTranscriptView: View {
     /// Drives keyboard focus into the rename field the moment it appears, so the
     /// user can type immediately instead of having to click the field first.
     @FocusState private var speakerFieldFocused: Bool
+    /// Explicit sidebar visibility, bound so our own toolbar toggle stays in a
+    /// fixed spot. Without this, NavigationSplitView supplies its own collapse
+    /// button, which relocates to the trailing edge of the title bar when the
+    /// sidebar is collapsed instead of staying next to Start Meeting.
+    @State private var columnVisibility: NavigationSplitViewVisibility = .all

     // MARK: - Displayed transcript

@@ -46,31 +51,61 @@ struct MeetingTranscriptView: View {
     private var isArchived: Bool { history.isShowingArchived }

     var body: some View {
-        NavigationSplitView {
+        NavigationSplitView(columnVisibility: $columnVisibility) {
             MeetingHistorySidebar()
                 .navigationSplitViewColumnWidth(min: 190, ideal: 210, max: 280)
+                // Removes the split view's automatic collapse button. This has to
+                // be done on the sidebar column itself; applied to the split view
+                // as a whole it has no effect. Ours replaces it, pinned in the
+                // leading toolbar group so it never moves.
+                .toolbar(removing: .sidebarToggle)
         } detail: {
             detailPane
         }
-        // Start/Stop lives in the toolbar, next to the sidebar-collapse button,
-        // so it stays reachable from anywhere (live or archived) at the very top.
+        // Start/Stop lives in the toolbar, next to our own sidebar toggle, so
+        // both stay reachable from anywhere (live or archived) at the very top.
+        // Unlike the system-supplied collapse button, the toggle stays put
+        // instead of jumping to the trailing edge once the sidebar is collapsed.
         .toolbar {
+            ToolbarItem(placement: .navigation) {
+                Button {
+                    columnVisibility = columnVisibility == .detailOnly ? .all : .detailOnly
+                } label: {
+                    Label("Toggle Sidebar", systemImage: "sidebar.left")
+                }
+                .accessibilityLabel("Toggle sidebar")
+            }
             ToolbarItem(placement: .navigation) {
                 Button {
                     Task { await meetingState.toggleMeeting() }
                 } label: {
-                    Label(
-                        meetingState.meetingState == .recording ? "Stop" : "Start Meeting",
-                        systemImage: meetingState.meetingState == .recording
-                            ? SFSymbols.stopFill
-                            : SFSymbols.recordingMicrophone
+                    HStack(spacing: 6) {
+                        Image(
+                            systemName: meetingState.meetingState == .recording
+                                ? SFSymbols.stopFill
+                                : SFSymbols.recordingMicrophone
+                        )
+                        .font(.body)
+
+                        Text(meetingState.meetingState == .recording ? "Stop" : "Start Meeting")
+                            .font(.callout.weight(.medium))
+                    }
+                    .padding(.horizontal, 12)
+                    .padding(.vertical, 6)
+                    .background(
+                        meetingState.meetingState == .recording
+                            ? Color.red.opacity(0.15)
+                            : theme.accentColor.opacity(0.15)
                     )
-                    .labelStyle(.titleAndIcon)
+                    .clipShape(RoundedRectangle(cornerRadius: 8, style: .continuous))
                 }
-                .buttonStyle(.borderedProminent)
-                .tint(meetingState.meetingState == .recording ? .red : theme.accentColor)
+                .buttonStyle(.plain)
                 .accessibilityLabel(
                     meetingState.meetingState == .recording ? "Stop meeting" : "Start meeting")
+                // Separates Start/Stop from the sidebar toggle so the two do not
+                // read as one control group. Sits outside the tinted shape, so it
+                // adds visual space without enlarging the tint itself.
+                .padding(.leading, 8)
             }
         }
         // Sized for sidebar + transcript: the rows spend ~154pt on the timestamp
```

## How I checked it

Release build of `e8fd62c` with the diff applied. Rather than trusting my eyes on the placement question, I read the window's accessibility tree, since that answers "did it actually move" objectively. `toolbar 1` ends up with exactly two buttons, and their coordinates are identical across three collapse/expand cycles:

```
toggle 1 -> buttons: Toggle Sidebar, Start Meeting | pos: 82,104  126,104
toggle 2 -> buttons: Toggle Sidebar, Start Meeting | pos: 82,104  126,104
toggle 3 -> buttons: Toggle Sidebar, Start Meeting | pos: 82,104  126,104
```

Two buttons rather than three confirms the system toggle is gone, and the fixed x-coordinates confirm nothing moves.

Item 1 is the one I'd class as a bug. Items 2 and 3 are judgement calls, so take or leave them, and feel free to put the toggle to the right of Start/Stop instead if you prefer that order. The 8pt gap is easy to tune too.

sebsto:
Will merge this and apply https://github.com/sebsto/wispr/pull/90 on top of it

PR #90, opened 2026-08-23
feat(meeting): browse, name, and manage saved transcripts

gbrunoo:
## Summary

Saved meeting transcripts were already being written to disk, but there was no way back to them from the app. This adds a history sidebar to the meeting window, plus the naming and cleanup operations that make an archive usable.

## What's in it

- **History sidebar** (`MeetingHistorySidebar`) listing saved sessions newest-first, with open / rename / delete.
- **`MeetingHistoryStore`** owns the list and the current selection. Kept deliberately separate from `MeetingStateManager` so that browsing history can never disturb a recording in progress, and starting a meeting can't wipe an archive the user is reading.
- **`TranscriptStore`** gains `list` / `load` / `delete` / `rename` alongside the existing `save`. Titles live inside the JSON and are only *slugged* into the filename, so the timestamp stays authoritative and filename and content can't drift into disagreement. Nothing is ever read back out of a filename.
- **Speaker naming**, keyed by diarization index, offered on finished sessions (which is when you actually know who "Speaker 2" was). Names flow through to copy and export, so an exported transcript reads the same as what's on screen.
- **`recordedSpan`** describes a saved session by its first-to-last timestamp. The existing `duration` is measured against *now*, which is correct while recording but grows forever for a session loaded from disk.
- **Backward-compatible decoding.** `MeetingTranscript` now decodes explicitly rather than by synthesis: the synthesized initializer requires every key to be present and does not fall back to a property's default, so adding `title` and `speakerNames` would otherwise fail to decode every transcript written before this change.
- A stray `Divider()` in the Meeting settings section is dropped. Inside a `Form` section macOS lays it out as its own inset row, so it read as an empty row above **Saved Transcripts** rather than as a separator.

## Bug fix included

`applicationWillTerminate` called `cancelRecording()`, which discards the transcript, so quitting mid-meeting destroyed the session with no warning. `stopMeeting()` can't be used there because it is `async` and the process is torn down before the suspension would resume, so `finalizeForTermination()` writes the session inline first and then cancels.

Happy to split that out into its own PR if you'd rather land it independently.

## Testing

- 19 new tests in `TranscriptHistoryTests`, covering listing order, load / delete, rename and filename slugging, speaker-name persistence and per-index keying, decoding of transcripts written before names existed, and span formatting.
- Full suite: 553 tests, all passing except `AudioEngine audio level stream terminates on stop`. That one is pre-existing on `main` — I confirmed by running the full suite there and getting the same single failure — and it passes in isolation on both branches, so it looks like microphone contention under the parallel run.
- `swift build --target WisprApp` clean.
- Also verified as a signed Release build installed and running locally on macOS 26.

### DIFF
--- a/Sources/WisprApp/Services/MeetingHistoryStore.swift
+++ b/Sources/WisprApp/Services/MeetingHistoryStore.swift
@@ -1,0 +1,148 @@
+//
+//  MeetingHistoryStore.swift
+//  wispr
+//
+//  Observable state for the meeting history sidebar: which sessions exist,
+//  which one is on screen, and the rename/delete operations on them.
+//
+
+import Foundation
+import Observation
+import os
+
+/// What the transcript pane is currently showing.
+enum TranscriptSelection: Hashable, Sendable {
+    /// The session being recorded now, or the last one recorded this launch.
+    case live
+    /// A session loaded from disk.
+    case archived(URL)
+}
+
+/// Owns the list of saved transcripts and the current selection.
+///
+/// Kept separate from `MeetingStateManager` so that browsing history never
+/// touches recording state: opening an old session must not be able to disturb
+/// a meeting that is running.
+@MainActor
+@Observable
+final class MeetingHistoryStore {
+
+    // MARK: - Published State
+
+    /// Saved sessions, newest first.
+    private(set) var summaries: [TranscriptSummary] = []
+
+    /// Which session the transcript pane shows.
+    private(set) var selection: TranscriptSelection = .live
+
+    /// The decoded transcript for an archived selection; `nil` when showing live.
+    private(set) var loadedTranscript: MeetingTranscript?
+
+    /// Surfaced in the sidebar when an operation fails, rather than failing silently.
+    var errorMessage: String?
+
+    /// Whether an archived session is on screen, as opposed to the live one.
+    var isShowingArchived: Bool {
+        if case .archived = selection { return true }
+        return false
+    }
+
+    /// Summary for the current archived selection, if any.
+    var selectedSummary: TranscriptSummary? {
+        guard case .archived(let url) = selection else { return nil }
+        return summaries.first { $0.url == url }
+    }
+
+    // MARK: - Loading
+
+    /// Re-reads the transcripts directory.
+    ///
+    /// If the open archived session has disappeared underneath us (deleted in
+    /// Finder, say), the selection falls back to live rather than leaving a
+    /// stale transcript on screen.
+    func refresh() {
+        summaries = TranscriptStore.list()
+
+        if case .archived(let url) = selection,
+            !summaries.contains(where: { $0.url == url })
+        {
+            showLive()
+        }
+    }
+
+    // MARK: - Selection
+
+    /// Shows the live session.
+    func showLive() {
+        selection = .live
+        loadedTranscript = nil
+    }
+
+    /// Loads and shows a saved session, staying put if it cannot be read.
+    func show(_ summary: TranscriptSummary) {
+        guard let transcript = TranscriptStore.load(summary.url) else {
+            errorMessage = "Could not open \(summary.displayName)."
+            return
+        }
+        loadedTranscript = transcript
+        selection = .archived(summary.url)
+        errorMessage = nil
+    }
+
+    // MARK: - Mutating
+
+    /// Names a session, or reverts it to its date when passed `nil`.
+    func rename(_ summary: TranscriptSummary, to title: String?) {
+        guard let newURL = TranscriptStore.rename(summary.url, to: title) else {
+            errorMessage = "Could not rename \(summary.displayName)."
+            return
+        }
+
+        let wasSelected = (selection == .archived(summary.url))
+        refresh()
+
+        // The URL is the row's identity and changes with the slug, so an open
+        // session is re-pointed at its new file instead of closing.
+        if wasSelected, let moved = summaries.first(where: { $0.url == newURL }) {
+            show(moved)
+        }
+        errorMessage = nil
+    }
+
+    /// Deletes a session. Irreversible: wispr keeps no audio to re-transcribe
+    /// from, so the caller is expected to have confirmed with the user.
+    func delete(_ summary: TranscriptSummary) {
+        guard TranscriptStore.delete(summary.url) else {
+            errorMessage = "Could not delete \(summary.displayName)."
+            return
+        }
+
+        if selection == .archived(summary.url) {
+            showLive()
+        }
+        refresh()
+        errorMessage = nil
+    }
+
+    /// Names a speaker in the session currently on screen, writing it straight
+    /// back to that session's file.
+    ///
+    /// Only meaningful for an archived session: the live transcript is owned by
+    /// `MeetingStateManager` and has no file to update until it is saved, which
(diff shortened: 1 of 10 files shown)


Sources/WisprApp/Services/MeetingStateManager.swift
```diff
@@ -193,11 +197,29 @@ final class MeetingStateManager {
...
 
+    /// Persists the transcript, then cancels recording.
+    ///
+    /// Called from `applicationWillTerminate`. `stopMeeting()` cannot be used
+    /// there: it is `async`, and the process is torn down before the suspension
+    /// would resume, so the session has to be written inline first. Without
+    /// this, closing the window mid-meeting and quitting from the menu destroys
+    /// the whole session, since `cancelRecording()` saves nothing.
+    func finalizeForTermination() {
+        lastSavedTranscriptURL = TranscriptStore.save(transcript)
```
sebsto, line 219: **[Correctness] `finalizeForTermination()` duplicates an already-saved (renamed) session.** — *Medium*

The live `transcript` is only reset by `startMeeting()`, so after a normal `stopMeeting()` (which already saved it) the entries remain in memory. On quit this re-saves them:

- Not renamed → `save(_:)` derives `meeting-<startTime>.json` and atomically overwrites the same file (redundant, harmless).
- **Renamed** → `TranscriptStore.rename` *moved* the file to `meeting-<startTime>-<slug>.json` and removed the original path. The re-save re-creates `meeting-<startTime>.json`, so history shows **two** rows for one session: the titled one and a fresh title-less copy with identical content.

The method's own doc says its purpose is to rescue an *in-progress* meeting that `cancelRecording()` would otherwise discard. Guarding on the recording state matches that intent and removes the duplicate:

```suggestion
    func finalizeForTermination() {
        // Only an in-progress meeting needs rescuing here: a stopped session was
        // already saved by stopMeeting(), and re-saving it re-creates a title-less
        // duplicate of a session the user may since have renamed.
        guard meetingState == .recording else { return }
        lastSavedTranscriptURL = TranscriptStore.save(transcript)
        cancelRecording()
    }
```

Sources/WisprApp/Services/MeetingHistoryStore.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    // MARK: - Loading
+
+    /// Re-reads the transcripts directory.
+    ///
+    /// If the open archived session has disappeared underneath us (deleted in
+    /// Finder, say), the selection falls back to live rather than leaving a
+    /// stale transcript on screen.
+    func refresh() {
+        summaries = TranscriptStore.list()
```
sebsto, line 64: **[Concurrency / Performance] `refresh()` fully decodes every transcript on the main actor.** — *Medium*

`refresh()` is `@MainActor` and calls `TranscriptStore.list()` synchronously — from `.task { history.refresh() }` and from every `.onChange(of: lastSavedTranscriptURL)`. `list()` calls `load(url)` for every file (`TranscriptStore.swift:131`), doing a full `Data(contentsOf:)` + JSON decode of each transcript and then discarding all but a few summary fields. That defeats the stated purpose of `TranscriptSummary` ("so the history list can be rendered without holding every full transcript in memory") and blocks the main actor for as long as the whole archive takes to read — unbounded as history grows.

The repo already keeps heavy work off the responsive path: `MeetingStateManager.startMeeting()` / `warmUpDiarizerIfEnabled()` run in `async` methods via `withTaskGroup` rather than inline on the actor. Same treatment here — do the read + decode off the main actor and hand back only the summaries:

```suggestion
    func refresh() async {
        let listed = await Task.detached { TranscriptStore.list() }.value
        summaries = listed

        if case .archived(let url) = selection,
            !summaries.contains(where: { $0.url == url })
        {
            showLive()
        }
    }
```

(Call sites become `.task { await history.refresh() }` and `.onChange(...) { Task { await history.refresh() } }`.) Decoding only the header fields is a lighter alternative, but moving the work off-actor is the smaller change and follows the existing pattern.

Sources/WisprApp/Services/TranscriptStore.swift
```diff
@@ -72,4 +82,212 @@ enum TranscriptStore {
...
+    var id: URL { url }
+
+    /// The user's name for the session, or its date when unnamed.
+    var displayName: String {
+        if let title, !title.isEmpty { return title }
+        return Self.dateFormatter.string(from: startTime)
+    }
+
+    /// `span` as "h:mm:ss", hours omitted when zero.
+    var formattedSpan: String {
```
sebsto, line 276: **[Maintainability] `formattedSpan` duplicates the span formatter verbatim.** — *Low (nit)*

This is byte-identical to `MeetingTranscript.formattedRecordedSpan` (`MeetingTranscript.swift:327`) and a near-copy of the pre-existing `MeetingTranscript.formattedDuration` (`MeetingTranscript.swift:307`) — three copies of the same h/mm/ss rule, which will drift. Since `TranscriptSummary` already carries the raw `span`, format through a single shared helper on `MeetingTranscript` instead of re-implementing it, so the "hours omitted when zero" logic lives in one place.

sebsto (commented):
**Review of `feat(meeting): browse, name, and manage saved transcripts`** — one COMMENT review.

Solid, well-scoped, and unusually well-documented. It turns already-on-disk transcripts into a browsable archive: a history sidebar, `list`/`load`/`delete`/`rename` on `TranscriptStore`, per-index speaker naming, an archive-safe `recordedSpan`, and a back-compat custom `Decodable`. Keeping `MeetingHistoryStore` separate from `MeetingStateManager` (so browsing can't disturb a live recording) is the right call and is honoured throughout `MeetingTranscriptView` via `displayedTranscript`.

I verified the headline back-compat concern directly and it is **correct**: the custom `init(from:)` covers all four `CodingKeys` (`entries`, `startTime`, `title`, `speakerNames`), decodes the two new keys with `decodeIfPresent` + sane defaults, and round-trips against the pre-existing `.iso8601` encoder. The legacy-decode test fixture matches the exact old on-disk shape. No forgotten or mis-defaulted key. Test isolation (real `TranscriptStore.directory` + `defer` cleanup) matches the existing `wisprTests/TranscriptStoreTests.swift` pattern, so it's consistent — not flagged.

Two things worth addressing before it lands, both **Medium**, neither blocking the design:

1. **`finalizeForTermination()` can create a duplicate archive file.** The live `transcript` is only reset by `startMeeting()`, so after a normal `stopMeeting()` the entries are still in memory; quitting re-saves them. If the session was renamed in between, `rename` has *moved* the file to a slugged name, so the re-save re-creates the original `meeting-<startTime>.json` — leaving two history rows for one session. Inline fix suggested.
2. **`TranscriptStore.list()` fully decodes every transcript on the `@MainActor`.** `MeetingHistoryStore.refresh()` calls it synchronously from `.task` and every `.onChange`; each entry is a full JSON decode whose result is thrown away except for a few summary fields — defeating the very reason `TranscriptSummary` exists, and blocking the main actor as history grows. Inline fix suggested.

Plus one Low nit (span-formatter triplication). Privacy: no transcript or speaker text is sent off-device or logged non-privately — only redacted filenames/error strings reach `Log`. CI `build-spm` + `build-xcode` are green on this HEAD; the `AudioEngine ... terminates on stop` flake is pre-existing on `main` (context, not this PR). Details inline.

Issue #98, opened 2026-08-26
SwiftUI review: lifecycle bugs, observation granularity, and cleanup punch list


sebsto:
A SwiftUI best-practices review of everything under `Sources/WisprApp/UI/` turned up a handful of real bugs and some anti-patterns. The codebase is in good shape overall — no deprecated APIs, no GCD/Combine/Timer, stable `ForEach` identity everywhere, Reduce Motion respected consistently — so this is a punch list, not a rewrite. Items are ordered by importance; each one says what's wrong, where, and how to fix it.

### 1. Closing Settings while recording a hotkey kills the global hotkey until relaunch

`SettingsView.swift:147-163` unregisters the global hotkey when hotkey recording starts (`isRecordingHotkey` becomes true) and only re-registers it when the flag flips back to false. But the Settings window is closable and `MenuBarController.openSettings()` builds a fresh `SettingsView` on every open. If the user closes the window mid-recording, the `@State` is thrown away, the flip-to-false never happens, and the dictation hotkey — the app's core trigger — stays dead until the user edits the hotkey or restarts the app.

**Fix:** make registration ownership single-sourced. Best option: move suspend/resume into `HotkeyMonitor` itself (e.g. `beginRecordingSession()` / `endRecordingSession()`, where dropping the session restores registration). Minimal option: add `.onDisappear` in `SettingsView` that re-registers if `isRecordingHotkey` was true.

### 2. Disabling the "Start/Stop Recording" menu item does nothing

`MenuBarController.swift:302` sets `recordingMenuItem.isEnabled = false` during processing, but `NSMenu.autoenablesItems` defaults to `true`, and under auto-enabling AppKit ignores manual `isEnabled` — it enables any item whose target responds to the action. `MenuBarActionHandler` doesn't implement `NSMenuItemValidation`, so the item is always clickable and the "don't allow toggling while processing" guard never engages.

**Fix:** either set `menu.autoenablesItems = false` in `buildMenu()` (around line 177), or implement `func validateMenuItem(_:) -> Bool` on `MenuBarActionHandler` returning `stateManager.appState != .processing` for that action.

### 3. Recording overlay can silently fail to appear (show/dismiss race)

[...]

PR #99, opened 2026-08-30
fix(menu): preserve disabled recording action while processing

hCoureau:
## Summary
- disable AppKit automatic menu-item enablement
- preserve the explicit disabled state of Start/Stop Recording while transcription is processing
- add a focused regression test for the menu behavior

## Testing
- Debug build with Xcode
- Focused AppKit regression check

### DIFF
--- a/Sources/WisprApp/UI/MenuBarController.swift
+++ b/Sources/WisprApp/UI/MenuBarController.swift
@@ -38,7 +38,7 @@ final class MenuBarController {
     private let statusItem: NSStatusItem
 
     /// The dropdown menu displayed when the user clicks the status item.
-    private let menu: NSMenu
+    let menu: NSMenu
 
     /// Reference to the central state manager for wiring actions.
     private let stateManager: StateManager
@@ -93,7 +93,7 @@ final class MenuBarController {
 
     // MARK: - Menu Items (retained for dynamic updates)
 
-    private let recordingMenuItem = NSMenuItem()
+    let recordingMenuItem = NSMenuItem()
     private let meetingMenuItem = NSMenuItem()
     private let stopMeetingMenuItem = NSMenuItem()
     private let languageMenuItem = NSMenuItem()
@@ -175,7 +175,7 @@ final class MenuBarController {
     /// Requirement 5.3: Menu contains Start/Stop Recording, Settings,
     /// Model Management, Language Selection, and Quit.
     private func buildMenu() {
-        Self.disableAutomaticItemEnablement(in: menu)
+        menu.autoenablesItems = false
         menu.removeAllItems()
 
         // Start/Stop Recording
@@ -277,15 +277,10 @@ final class MenuBarController {
         }
     }
 
-    /// Preserves explicit item enablement, such as disabling recording while processing.
-    static func disableAutomaticItemEnablement(in menu: NSMenu) {
-        menu.autoenablesItems = false
-    }
-
     // MARK: - Recording Menu Item
 
     /// Updates the recording menu item title and action based on current state.
-    private func updateRecordingMenuItem() {
+    func updateRecordingMenuItem() {
         let isRecording = stateManager.appState == .recording
         let shortcut = KeyCodeMapping.shared.hotkeyDisplayString(
             keyCode: settingsStore.hotkeyKeyCode,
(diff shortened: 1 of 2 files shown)


sebsto:
Hello Harry @hCoureau 

Thank you for proposing this PR.
Can you explain in human readable words what is the problem this PR fixes ?

wisprTests/MenuBarControllerTests.swift
```diff
@@ -136,6 +136,25 @@ struct MenuBarControllerIconTests {
 @Suite("MenuBarController Menu Structure Tests")
 struct MenuBarControllerMenuTests {
 
+    @Test("Menu preserves an explicitly disabled recording action")
+    func testMenuPreservesExplicitDisabledState() {
```
sebsto: **[Tests]** This exercises a throwaway `NSMenu` + the static helper and asserts AppKit honours `isEnabled = false` — i.e. it tests `NSMenu`'s documented `autoenablesItems` semantics, not wispr's wiring. It never goes through `buildMenu()` or touches the real `recordingMenuItem`, so it would still pass if someone deleted the `Self.disableAutomaticItemEnablement(in: menu)` call (line 178) or the `recordingMenuItem.isEnabled = …` line (line 308) — the two lines that actually implement #98.

Stronger: relax `recordingMenuItem` from `private` to internal (reachable via this file's `@testable import WisprApp`) and drive the real controller, the way every other test in this file already does via `createTestController()` (line 20, used e.g. at line 160):

```swift
@Test("Recording item is disabled while processing")
func testRecordingItemDisabledWhileProcessing() async {
    let (controller, stateManager, _, _) = createTestController()
    stateManager.appState = .processing   // or the real transition
    controller.updateRecordingMenuItemForTest()   // or observe
    #expect(!controller.recordingMenuItem.isEnabled)
    controller.stopObserving()
}
```

Minor: give the assertion an explanatory message like the rest of this file (e.g. line 99).

Sources/WisprApp/UI/MenuBarController.swift
```diff
@@ -276,6 +277,11 @@ final class MenuBarController {
         }
     }
 
+    /// Preserves explicit item enablement, such as disabling recording while processing.
+    static func disableAutomaticItemEnablement(in menu: NSMenu) {
```
sebsto: **[Simplicity]** This static method wraps a single property assignment, and its only reason to exist is letting the test reach the behaviour without a controller. If the test moves to the real-menu approach (see my note on the test file), that justification goes away — drop the method and set the flag directly at the top of `buildMenu()`:

```swift
private func buildMenu() {
    menu.autoenablesItems = false
    menu.removeAllItems()
    ...
}
```

Non-blocking; only worth doing alongside the test change.

sebsto (commented):
Fixes #98 item 2 correctly and minimally. Under the default `autoenablesItems = true`, AppKit ignored the manual `recordingMenuItem.isEnabled = false` and kept Start/Stop Recording clickable during `.processing`; flipping the flag off makes that explicit disable stick.

**Footgun check — does turning auto-enablement off strand other items?** No. Line 308 is the only `isEnabled` write in the whole controller, so every other item keeps its default `isEnabled = true`; nothing relied on auto-*disable* (all items target `MenuBarActionHandler.shared`, which responds to every selector), and the dynamic items use `isHidden`, not `isEnabled`. `languageSubmenu` is a separate `NSMenu`, so its own enablement is untouched. Re-enable after processing is intact — `startObservingState()` re-runs `updateRecordingMenuItem()` whenever `appState` leaves `.processing` (complete / fail / cancel).

On the design choice: #98 also floated overriding `validateMenuItem(_:)`. The `autoenablesItems = false` route you took is the better fit here — this menu is already manually managed through retained items and `updateRecordingMenuItem()`, so keeping enablement in one place beats splitting it into a validate method. No change wanted there.

The two notes below are both about the regression test's depth, not the fix itself. Neither blocks.

(CI on the Mac runners is green; I can't compile Apple code on my host, so that's context, not a build claim.)

hCoureau:
Thanks — agreed. The current test verifies AppKit semantics but not Wispr wiring, so it would not catch a future removal of the buildMenu configuration or the explicit processing-state disable. I will switch it to exercise the real controller and recording menu item, add an explanatory assertion message, and fold the one-line helper back into buildMenu.

hCoureau:
Updated in dfb01c5. The helper is removed and the flag now lives directly in buildMenu(). The regression test drives the real MenuBarController into processing, runs AppKit validation on its real menu, and asserts the retained recording item remains disabled with an explanatory message.

sebsto:
Nice — this is exactly what I was hoping for. The test now goes through the real controller instead of a stand-in menu, so it'll actually catch it if the disabled-while-processing behavior ever breaks again. And dropping the static helper for a plain `menu.autoenablesItems = false` reads cleaner too.

Thanks for the quick turnaround, @hCoureau 🙏

hCoureau:
No Problem! 🎉

PR #100, opened 2026-08-31
feat(meeting): in-person mode with per-speaker mic diarization

### LINKED ISSUE #97: Meeting mode: in-person recording with per-speaker separation
## The idea

Meeting mode assumes a *remote* meeting: your voice on the mic, everyone else's through system audio. That shape is baked in from capture all the way to the transcript labels.

For an in-person meeting — everyone around one table, nothing playing through the speakers — that model breaks down. All voices arrive on the mic, so today the whole room transcribes as a single speaker, `You`.

Proposal: an **In person / Online** switch on the meeting window, where in-person diarizes the *microphone* track and labels each person individually, with no privileged "you".

Opening this to agree the shape before writing code, since a few of the decisions below are yours to make rather than mine.

## What in-person actually implies

Worth being precise about one thing, because it changes the data model rather than just the display: in an in-person meeting **you are one of the participants**, and your voice reaches the same mic as everyone else's. So this isn't "hide `You`" — it's "`You` stops being a distinct concept", and the local speaker becomes just another diarized index. Arguably an improvement (you can name yourself), but it's a model change, not a filter.

## The blocking constraint

Diarization currently runs on the system-audio track only. `MeetingStateManager.transcribeSystemAudio()` calls `diarizer.ingest(_:at:)` and `dominantSpeaker(in:)`; `transcribeMicAudio()` doesn't involve the diarizer at all and hardcodes `speaker: .you`.

The two capture paths also aren't symmetrical, and this is the real prerequisite:

| stream | type |
| --- | --- |
| `MeetingAudioEngine.systemAudioStream` | `AsyncStream<SystemAudioChunk>` — carries `startTime` |
| `MeetingAudioEngine.micAudioStream` | `AsyncStream<[Float]>` — **no timestamps** |

`MeetingDiarizer` needs a timeline for both `ingest(_:at:)` and `dominantSpeaker(in: ClosedRange<TimeInterval>)`. So diarizing the mic requires giving the mic path a timestamped chunk type mirroring `SystemAudioChunk`. That's an audio-pipeline change in `MeetingAudioEngine`, and it's the bulk of the work — the UI toggle is the easy part.

Suggestion: land the timestamped mic chunks as their own PR first, so the risky pipeline change gets reviewed on its own rather than buried under UI. It's invisible to users and independently reviewable.

One thing in our favour: the mic-only path already exists as a graceful fallback (system capture failure logs a warning and continues mic-only), so in-person is closer to "deliberately take the existing fallback" than to new capture plumbing.

## Modelling options

`MeetingSpeaker` today is `.you` / `.others(speakerIndex: Int?)`.

**A. Add a case** — `case participant(index: Int)`. Most honest, but every `switch` grows a branch and `displayName` / `speakerIndex` / `isRemote` all have to answer for it.

**B. Reuse `.others(speakerIndex:)` for everyone in-person, and record the mode on `MeetingTranscript`.** No new case, `speakerNames` keeps keying by index unchanged, and `displayName` consults the mode to decide between "Speaker 2" and "Others". Nothing in-person ever produces `.you`. Smallest diff; slight semantic lie in the name "others" containing you.

**C. Make the names mode-neutral** — `.you`/`.others` → something like `.local`/`.tracked(index:)`. Cleanest long-term, biggest blast radius, and it churns the on-disk `Codable` shape.

I'd start with **B**, with the mode stored on `MeetingTranscript` (via `decodeIfPresent`, defaulting to online) so archived transcripts keep rendering correctly and older files still decode.

## Knock-on decisions

- **Diarization stops being optional.** It's `false` by default today and marked experimental, with a ~30 MB model on first use. In-person without it collapses the room into one speaker, which defeats the mode. Does the toggle silently enable it, or refuse until the model is present? Needs a story for "still downloading".
- **Echo suppression should be bypassed, not just defaulted off.** It exists to drop mic echo of remote speech (#65). In-person there is no remote audio, so it can only cause false positives — two people in a room saying similar short phrases is exactly the case the 3-word minimum guards against.
- **Capture system audio in-person at all?** Skipping it avoids the Screen Recording permission prompt entirely, which is a real UX win. But it forecloses **hybrid** meetings (people in the room plus a remote dial-in), which may be the most common case in practice. Explicit non-goal for v1, or a third mode later?
- **Mid-meeting switching.** Entries already recorded carry the old mode's labels. Simplest defensible answer: disable the toggle while recording, so a transcript never mixes two labelling schemes.
- **Diarizer cold start.** `dominantSpeaker` returns `nil` for the first chunk or two. Online that renders as "Others" and reads fine; in-person `nil` has no sensible label. Hold those entries briefly, show a placeholder that resolves, or accept one unlabelled opening line?

## Smaller UI items

- The header's level indicators are hardcoded `"You"` and `"Others"`; in-person only the mic level is meaningful.
- `speakerColor` gives `.you` blue and remote speakers a rotating palette. In-person nobody should be privileged — everyone on the palette.
- Naming already flows through `displayName(for:)` to copy and export, so named participants should need no extra work there.

## Questions

1. Model A, B or C ?
2. Toggle in the meeting window header (per session) or Settings (global default)? Or header, remembering the last choice? 
3. Should "In person" enable diarization silently, or block with an explanation until the model is ready? - I'd go for displaying "others" while the model loads

Happy to take this on once the shape is agreed.

gbrunoo:
Add an Online/In person capture mode to the meeting window. In-person records the microphone only (no system audio, no Screen Recording permission) and diarizes the mic track so each person in the room becomes a numbered speaker, with no privileged "You".

- MeetingMode persisted on MeetingTranscript (decodeIfPresent, defaults to online so archived transcripts keep rendering)
- Timestamped MicAudioChunk so the diarizer has a mic timeline
- In-person forces diarization on and bypasses echo suppression (no remote audio to echo); labels reuse .others(speakerIndex:)
- Header mode toggle (locked while recording); in-person shows a single Room level; up to four participants (Sortformer limit)
- History sidebar now closed by default; opening it widens the panel and closing restores the prior compact width unless the user resized it
- Header controls never wrap: Start/Stop and the level indicator collapse to icon-only / hide when the window is too narrow

image
image
image

### DIFF
--- a/Sources/WisprApp/Models/MeetingTranscript.swift
+++ b/Sources/WisprApp/Models/MeetingTranscript.swift
@@ -7,6 +7,23 @@
 
 import Foundation
 
+/// How a meeting is captured, which determines the speaker model.
+///
+/// - `online`: the classic remote-meeting shape. Your voice is on the mic and
+///   everyone else arrives through system audio; the mic track is labelled
+///   "You" and only the system track is diarized.
+/// - `inPerson`: everyone is in one room and reaches the same microphone. There
+///   is no privileged "You" — the mic track itself is diarized and every
+///   participant (including the local user) becomes a numbered speaker. System
+///   audio is not captured, so no Screen Recording permission is needed.
+///
+/// Diarization can distinguish up to four participants in person, a fixed limit
+/// of the underlying Sortformer model.
+nonisolated enum MeetingMode: String, Sendable, Codable, CaseIterable {
+    case online
+    case inPerson
+}
+
 /// Identifies the audio source / speaker in a meeting transcript.///
 /// Speaker indices are stored **0-based** internally. The +1 transform to
 /// "Speaker 1", "Speaker 2", etc. happens only inside `displayName`.
@@ -69,6 +86,11 @@ nonisolated struct MeetingTranscript: Sendable, Equatable, Codable {
     var entries: [MeetingTranscriptEntry] = []
     let startTime: Date
 
+    /// How this session was captured. Fixed at `startMeeting()` and never changed
+    /// mid-session, so a transcript never mixes two labelling schemes. Archived
+    /// transcripts written before this field existed decode as `.online`.
+    var mode: MeetingMode = .online
+
     /// User-assigned title for the session, or `nil` to fall back to its date.
     ///
     /// A meeting's date is a poor handle once there are more than a handful of
@@ -91,7 +113,7 @@ nonisolated struct MeetingTranscript: Sendable, Equatable, Codable {
     // MARK: - Codable
 
     private enum CodingKeys: String, CodingKey {
-        case entries, startTime, speakerNames, title
+        case entries, startTime, speakerNames, title, mode
     }
 
     /// Decodes a transcript, tolerating files written before `speakerNames` and
@@ -109,6 +131,7 @@ nonisolated struct MeetingTranscript: Sendable, Equatable, Codable {
         self.speakerNames =
             try container.decodeIfPresent([String: String].self, forKey: .speakerNames) ?? [:]
         self.title = try container.decodeIfPresent(String.self, forKey: .title)
+        self.mode = try container.decodeIfPresent(MeetingMode.self, forKey: .mode) ?? .online
     }
 
     // MARK: - Title
--- a/Sources/WisprApp/Services/MeetingStateManager.swift
+++ b/Sources/WisprApp/Services/MeetingStateManager.swift
@@ -53,10 +53,13 @@ final class MeetingStateManager {
     /// Always 0 in `.inPerson` mode, which does not capture system audio.
     var systemLevel: Float = 0
 
-    /// Capture mode for the next (or current) meeting. Bound to the header
-    /// toggle and seeded from the last remembered choice. Editable only while
-    /// idle — the header disables the toggle during recording so a transcript
-    /// never mixes two labelling schemes.
+    /// Capture mode for the **next** meeting. Bound to the header toggle and
+    /// seeded from the last remembered choice.
+    ///
+    /// Once a meeting starts this is copied to `transcript.mode`, which is the
+    /// canonical mode for the running session — every mode-dependent decision
+    /// (labelling, diarization, echo suppression) reads that frozen value rather
+    /// than this binding, so a transcript can never mix two labelling schemes.
     ///
     /// Selecting in-person turns on diarization automatically: it is what
     /// separates the people in the room, and without it the whole room collapses
@@ -186,7 +189,9 @@ final class MeetingStateManager {
     /// collapses into a single speaker, which defeats the mode. Online mode only
     /// warms it when the user opted into diarization.
     private func warmUpDiarizerIfEnabled() async {
-        let wantsDiarization = mode == .inPerson || settingsStore.meetingDiarizationEnabled
+        // `transcript.mode` is the frozen session mode, set once at startMeeting().
+        let wantsDiarization =
+            transcript.mode == .inPerson || settingsStore.meetingDiarizationEnabled
         guard wantsDiarization, let diarizer = meetingDiarizer else {
             return
         }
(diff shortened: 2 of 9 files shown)


sebsto (commented):
Reviewed against the #97 design — this is a faithful **Model B** implementation, no silent divergence. Everyone in person becomes a numbered `.others(speakerIndex:)` with no privileged "You" (so `speakerColor`/`displayName` needed no change), `MeetingMode` rides on `MeetingTranscript` with a `decodeIfPresent` `.online` default, the timestamped `MicAudioChunk` mirrors `SystemAudioChunk` (the sample→`startTime` bookkeeping is line-for-line the same as the system path), diarization is forced on in person, echo suppression is bypassed, the header toggle is locked while recording, and `.inPerson` returns before touching ScreenCaptureKit so the Screen Recording prompt is genuinely avoided. Cold-start chunks fall back to `.others(nil)` → "Others", matching the answer to Q3.

Privacy/security is clean: no new network or telemetry, and the debug logs carry sample counts and timestamps, never audio or transcript text. Concurrency is tidy too — the diarizer is driven by exactly one path per mode, so there's no concurrent-ingest race, and the new AppKit callback captures `[weak self]`.

Everything below is **Low**. The one worth acting on is coverage: the three new behaviors (mode back-compat, the in-person label path, and the echo-suppression bypass) all ship without tests, and this repo normally tests exactly this kind of change.

(No build verification from my side — Apple-platform code; CI on the Mac runners is green.)

Sources/WisprApp/Models/MeetingTranscript.swift
```diff
@@ -109,6 +131,7 @@ nonisolated struct MeetingTranscript: Sendable, Equatable, Codable {
         self.speakerNames =
             try container.decodeIfPresent([String: String].self, forKey: .speakerNames) ?? [:]
         self.title = try container.decodeIfPresent(String.self, forKey: .title)
+        self.mode = try container.decodeIfPresent(MeetingMode.self, forKey: .mode) ?? .online
```
sebsto, line 134: **[Tests]** No back-compat decode test for `mode`. The repo already tests this exact class of change — `TranscriptStoreTests.swift:75` ("Legacy JSON without title decodes with no title") and `:31` for `speakerNames`. Worth the sibling test: a legacy transcript JSON with no `mode` key decodes as `.online`, and `.inPerson` survives an encode/decode round-trip. Cheap to add and it locks the archived-file guarantee in place.

Sources/WisprApp/Services/MeetingStateManager.swift
```diff
@@ -334,8 +399,12 @@ final class MeetingStateManager {
     /// setting is enabled. Echo suppression drops mic ("You") transcriptions that
     /// duplicate a recent remote ("Others") transcription — the result of remote
     /// speech leaking from the speakers into the mic (issue #65).
+    ///
+    /// In-person mode bypasses echo suppression entirely: there is no remote
+    /// audio, so the mechanism can only produce false positives (two people in a
+    /// room saying similar short phrases).
     private func record(_ entry: MeetingTranscriptEntry) {
-        guard settingsStore.meetingEchoSuppressionEnabled else {
+        guard mode == .online, settingsStore.meetingEchoSuppressionEnabled else {
```
sebsto: **[Maintainability]** This reads the live `mode` rather than the frozen `transcript.mode`. `transcript.mode` is documented (`MeetingTranscript.swift:89-92`) as "Fixed at `startMeeting()` and never changed mid-session" — it's the canonical session mode, and it's already set at start. Gating on it here (and in `transcribeMicAudio`/`warmUpDiarizerIfEnabled`) would make the echo-suppression bypass depend on that invariant instead of on the header staying locked. It's correct today only because the toggle is disabled while recording.

**[Tests]** The in-person bypass itself is uncovered — `MeetingEchoSuppressionTests` drive `appendSuppressingEcho` at the transcript level, not the `mode` gate in `record()`, so nothing asserts that in person a would-be echo is kept.

Sources/WisprApp/Services/MeetingStateManager.swift
```diff
@@ -267,17 +304,45 @@ final class MeetingStateManager {
...
+                    // nil (cold-start or diarization off) renders as plain "Others".
+                    var speakerIndex: Int?
+                    if let diarizer {
+                        let chunkEnd =
+                            chunk.startTime
+                            + Double(chunk.samples.count) / Double(MeetingAudioEngine.sampleRate)
+                        speakerIndex = await diarizer.dominantSpeaker(
+                            in: chunk.startTime...chunkEnd)
+                    }
+                    speaker = .others(speakerIndex: speakerIndex)
```
sebsto: **[Tests]** The in-person label path has no coverage. Nothing asserts that a resolved chunk yields `.others(index)` while a cold-start / diarization-off chunk yields `.others(nil)` (rendered "Others"), and never `.you`. `MeetingStateManagerTests.swift:103` ("MeetingSpeaker display names are correct") is the natural place to extend — this is the behavioural heart of the feature.

Sources/WisprApp/UI/Meeting/MeetingTranscriptView.swift
```diff
@@ -54,16 +62,25 @@ struct MeetingTranscriptView: View {
...
+            // mode picker to remain on one line. When history is open, the panel
+            // grows by exactly its fixed sidebar width rather than squeezing here.
             detailPane
+                .frame(minWidth: Self.compactMinimumWidth)
+        }
+        .frame(
+            minWidth: showHistory
+                ? Self.compactMinimumWidth + Self.historyWidthIncrement
+                : Self.compactMinimumWidth,
+            minHeight: 420
```
sebsto: **[Maintainability]** This PR nicely shares width through `compactMinimumWidth` / `historyWidthIncrement` so the panel and the view can't drift apart. The min-height is the odd one out: `420` is a bare literal here while `MeetingWindowPanel.minimumHeight` is a named `420` — the same value declared in two spots. Fold it into a shared constant the same way you did the widths.

```suggestion
            minHeight: Self.minimumHeight
```
(with a `static let minimumHeight: CGFloat = 420` alongside the width constants, referenced from the panel too.)

sebsto:
This is looking great. Gating labelling, diarizer warmup and echo suppression on the frozen `transcript.mode` is the right call — the mode can't drift mid-session now, so correctness doesn't lean on the toggle being locked. Pulling `micSpeaker(mode:diarizedIndex:)` out into something testable was a nice touch too, and the echo-suppression test with the online control makes it obvious what the bypass actually changes.

Everything from the last round is covered. Thanks for seeing it through, @gbrunoo — merging shortly. And thanks for building out the in-person mode from #97. 🙏

PR #102, opened 2026-09-01
feat(vocabulary): revive custom vocabulary correction

hCoureau:
## Summary

- Revives the custom vocabulary correction work from #78 on current `main`.
- Preserves the original implementation commits by @omansour (Olivier Mansour).
- Addresses Seb’s review feedback: accurate Parakeet auto-detect fallback documentation, `eau` phonetic folding order, short multi-word matching, and stable SwiftUI row identity.
- Carries vocabulary correction through dictation and meeting transcription paths.

## Attribution

Original feature implementation and follow-up review work: @omansour. This PR rebases and completes that contribution with the maintainer’s approval.

Supersedes / follows: #78

## Testing

- `xcodebuild build -project wispr.xcodeproj -scheme wispr -configuration Debug -destination 'platform=macOS,arch=arm64' CODE_SIGN_IDENTITY=- CODE_SIGN_STYLE=Manual DEVELOPMENT_TEAM=`
- `swift test --skip-build --filter VocabularyCorrectorTests`
- `swift test --skip-build --filter MeetingVocabularyCorrectionTests`
- Full `swift test` exercised the feature suites; its only failure was the existing timing-sensitive `AudioEngine audio level stream terminates on stop` test, outside this change.

### DIFF
--- a/Sources/WisprApp/Utilities/VocabularyCorrector.swift
+++ b/Sources/WisprApp/Utilities/VocabularyCorrector.swift
@@ -38,6 +38,13 @@ enum VocabularyCorrector {
     /// tokens ("data ecou"), so we must look at multi-word windows.
     private static let maxWindow = 3
 
+    /// Minimum phonetic-key length for fuzzy matching. On 1–2 character keys the
+    /// normalized edit distance is effectively binary, so the similarity threshold
+    /// cannot discriminate ("Go" → key "g" would match any word folding to "g").
+    /// Shorter keys only match when both keys are identical AND the original
+    /// words are equal case-insensitively.
+    private static let minKeyLengthForFuzzyMatch = 3
+
     /// A vocabulary entry paired with its phonetic key per candidate language.
     private struct Entry {
         let canonical: String
@@ -185,6 +192,14 @@ enum VocabularyCorrector {
             var entryScore = 0.0
             for language in languages {
                 guard let wk = windowKeys[language], let ek = entry.keysByLanguage[language] else { continue }
+                // Keys too short for the threshold to discriminate: only accept a
+                // case-insensitive spelling match (still fixes casing, e.g. "go" → "Go").
+                if ek.count < minKeyLengthForFuzzyMatch || wk.count < minKeyLengthForFuzzyMatch {
+                    if joined.caseInsensitiveCompare(entry.canonical) == .orderedSame {
+                        entryScore = 1
+                    }
+                    continue
+                }
                 entryScore = max(entryScore, similarity(wk, ek))
             }
             if entryScore >= similarityThreshold, entryScore > (best?.score ?? 0) {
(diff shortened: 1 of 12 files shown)


Sources/WisprApp/Utilities/VocabularyCorrector.swift
```diff
@@ -0,0 +1,431 @@
...
+        var end = chars.count
+        while start < end, !chars[start].isLetter, !chars[start].isNumber { start += 1 }
+        while end > start, !chars[end - 1].isLetter, !chars[end - 1].isNumber { end -= 1 }
+        let leading = String(chars[0..<start])
+        let trailing = String(chars[end..<chars.count])
+        return (leading, trailing)
+    }
+
+    /// The alphanumeric core of a token, with surrounding punctuation removed.
+    private static func core(of token: String) -> String {
```
sebsto: **[Simplicity]** `core(of:)` looks unused — every caller goes through `alphanumericCore(of:)` instead (the window join at `:183` and the short-key compare at `:198`). `affixes(of:)` covers the punctuation-splitting the reconstruction needs. Looks like leftover scaffolding from #78; safe to drop the whole method.

Sources/WisprApp/Utilities/VocabularyCorrector.swift
```diff
@@ -0,0 +1,431 @@
...
+            if ch != last { deduped.append(ch) }
+            last = ch
+        }
+
+        // Drop a trailing schwa-ish vowel that speakers/models often add or
+        // drop ("novaria" vs "novariou" → both end effectively the same).
+        while let lastChar = deduped.last, "aeiou".contains(lastChar), deduped.count > 1 {
+            deduped.removeLast()
+            break
+        }
```
sebsto: **[Style]** This loop always hits `break` on the first pass, so it only ever strips a single trailing vowel — which matches the comment (“a trailing schwa-ish vowel”, singular). The `while`+`break` reads as if it iterates; an `if` says exactly what it does.

```suggestion
        if let lastChar = deduped.last, "aeiou".contains(lastChar), deduped.count > 1 {
            deduped.removeLast()
        }
```

sebsto (commented):
Revives @omansour's #78 vocabulary-correction work on current `main`, and it closes out all four of the earlier review points — the `eau`-before-components fold ordering, short multi-word matching, the Parakeet / auto-detect English-fallback behaviour (documented and tested), and stable list-row identity — each with a matching test.

The `VocabularyCorrectorTests` suite is genuinely thorough: casing normalization, multi-token splits (`data ecou` → canonical), neighbour non-absorption at equal score, multi-language PyTorch, punctuation reattachment, start/end positions, and the `eau` ordering. The corrector is wired into both the dictation path (`StateManager.insertTranscribedText`) and the meeting path after AI correction, so it has the final say on proper-noun spelling, and it stays inert unless enabled with a non-empty list. Privacy is intact — purely local string work, no new network or dependencies.

Nothing blocking. Two small cleanups inline (both Low). CI is green on the Mac runners; I can't build Apple-platform code from my end, so that's context, not a build claim from me.

hCoureau:
Feedback integrated! Thanks!

PR #107, opened 2026-09-17
Fix clipboard state after failed text insertion

fbettag:
A failed paste leaves `originalPasteboardContents` set. If the user then copies something new and dictates again, the next successful insertion restores the stale snapshot, overwriting that newer copy.

Clear the saved snapshot and cancel any pending restore when either writing to the clipboard or posting the paste shortcut fails. A failed shortcut still retains the transcription for manual paste. Also check PostEvent access before posting the real shortcut so missing permission follows the failure path. The layout-aware key resolver is unchanged.

Validation: the new deterministic regression fails on main (restores “before failure” instead of “new user copy”) and passes with this fix. Both failure regressions were reproduced before their fixes. All six `TextInsertionClipboardTests` pass via `swift test --filter TextInsertionClipboardTests` with Xcode 26.0. An injected access predicate now exercises permission denial, verifies that no paste events are posted, preserves the manual fallback, and verifies that a later permission grant is rechecked. Actual macOS permission integration still needs GUI verification.


Sources/WisprApp/Services/TextInsertionService.swift
```diff
@@ -141,6 +142,11 @@ final class TextInsertionService: TextInserting {
         let success = performPaste()
 
         guard success else {
+            // A failed paste leaves text available for manual pasting, but must
+            // not retain a snapshot that a later insertion could restore.
+            pasteboardRestoreTask?.cancel()
+            pasteboardRestoreTask = nil
+            originalPasteboardContents = nil
```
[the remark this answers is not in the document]
fbettag: Fixed in 4d9a808: both failure paths now use the same cancellation and snapshot reset helper. Added a setString failure regression, reproduced it failing before the fix, and verified all five clipboard tests pass.

fbettag:
Addressed the latest review's permission-coverage request in fa688aa. The access predicate is injectable and checked on every insertion before posting events. The new regression covers denied access, no posted paste events, preserved manual fallback text, and a later permission grant without recreating the service. All six clipboard tests pass. This tests the application's denied-access behavior; actual macOS permission integration still requires GUI verification.

Sources/WisprApp/Services/TextInsertionService.swift
```diff
@@ -76,15 +77,22 @@ final class TextInsertionService: TextInserting {
...
+    /// Checked on each insertion so changes to macOS permissions take effect immediately.
+    private let canPostEvents: @MainActor () -> Bool
+
     // MARK: - Init
 
     init(
         pasteboard: any TextPasteboard = NSPasteboard.general,
         restoreDelay: Duration = .seconds(2),
+        canPostEvents: @escaping @MainActor () -> Bool = {
+            CGPreflightPostEventAccess() || AXIsProcessTrusted()
```
sebsto: **[Security — LOW, just a question, nothing to fix necessarily]** The default gate is `CGPreflightPostEventAccess() || AXIsProcessTrusted()`. Since the actual paste is posted with `CGEvent.post(tap: .cghidEventTap)`, which needs the *PostEvent* privilege specifically, a process that is Accessibility-trusted but does **not** have PostEvent access would pass this preflight and then rely on `performPaste()` failing for real. That's harmless (the real post stays the true guard, so it just becomes a redundant early-out), but I wanted to check the intent: is AX-trusted meant to be sufficient here, or would `CGPreflightPostEventAccess()` alone match what you're gating on? Totally fine to leave as-is if the `||` is deliberate — just flagging it so the reasoning is captured. 🙂

sebsto (commented):
Thanks so much for this contribution, Franz — and welcome! 🙌 This is a really clean first PR: a well-scoped fix, a clear write-up of the bug, and genuinely thorough tests. Nice work.

The core fix is correct. Routing both failure paths through `discardPendingPasteboardRestore()` (cancel the pending task, drop the task handle, clear `originalPasteboardContents`) properly closes the stale-snapshot bug — a later insertion can no longer restore a snapshot left behind by a failed paste. It also lines up nicely with the existing invariant that only the *active* restore task clears the shared state.

On security (which I looked at closely): nothing concerning. The `canPostEvents()` preflight is non-prompting and runs on `@MainActor`, so it won't pop a permission dialog on every insertion; `success = canPostEvents() && performPaste()` short-circuits so no key events are posted when access is denied (your test asserts `pasteCalls == 0` — good); and no transcription content is logged. I left one small, non-blocking question inline about the `||` in the default gate.

One thing worth noting for future readers (no change needed): on a denied paste you intentionally leave the transcription on the clipboard for manual pasting, and I confirmed `StateManager`'s catch block re-copies it and tells the user, so the "restore within 2s" privacy note simply doesn't apply on the failure path — that's by design.

CI is green on all three legs (build-spm, build-xcode, test-spm) and it's mergeable. Really solid — thank you again for taking the time to write the regression tests. 🎉

Sources/WisprApp/Services/TextInsertionService.swift
fbettag: Thanks for questioning this. I removed the AX fallback in 1c87aa6: CGPreflightPostEventAccess alone now gates the operation we actually perform. The SDK also declares CGEventPost as void, so performPaste cannot detect OS rejection after posting; it only detects event-construction failures. That makes the narrower preflight preferable to an unverified AX fallback. All six clipboard regressions pass, including denied access and a later grant. Actual GUI/TCC integration remains unverified for this build.

PR #109, opened 2026-09-17
Add Right Option as a configurable dictation hotkey

fbettag:
Users can select bare Right Option in the hotkey recorder and use it for dictation through the existing modifier-key event tap. The default shortcut remains unchanged.

The handler distinguishes physical Right Option from Left Option using device-specific flags. It emits one press/release pair, ignores presses made while Command, Control, Shift, Fn, or Left Option is already held, and still emits release if another modifier is held when Right Option is released. Releasing another modifier while Right Option remains held does not start a new dictation. Caps Lock is intentionally allowed in both shortcut recording and dictation because it is a latched state. Existing Fn handling and Carbon shortcuts are preserved.

Like other immediate modifier-only hotkeys, pressing bare Right Option starts dictation immediately; this does not defer activation to distinguish a later Option+letter chord. Users who need Right Option for text entry can keep their existing shortcut.

Validation:
- Built with Xcode 26.0 / Swift 6.2 using SwiftPM.
- `swift test --filter 'RightOptionHotkeyTests|HotkeyMonitorTests'`: 25 tests passed, including parameterized modifier combinations, Left Option, Caps Lock, repeated events, release while another modifier is held, unregister/reset, and display naming.
- Tests exercise the event state machine directly. Physical keyboard delivery and selecting the shortcut in the running GUI have not been manually verified for this upstream build.


Sources/WisprApp/Services/HotkeyMonitor.swift
```diff
@@ -416,6 +428,29 @@ final class HotkeyMonitor {
...
+        rightOptionIsDown = isDown
+
+        if !isDown {
+            guard rightOptionDidTrigger else { return false }
+            rightOptionDidTrigger = false
+            onHotkeyUp?()
+            return true
+        }
+
+        let otherModifiers: CGEventFlags = [.maskCommand, .maskControl, .maskShift, .maskSecondaryFn]
```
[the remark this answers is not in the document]
fbettag: Caps Lock is intentionally allowed: its latched state should not prevent dictation. Clarified the PR description and handler documentation, and added parameterized press/release coverage with Caps Lock enabled in 37970e5. All 25 hotkey tests pass.

Sources/WisprApp/UI/Settings/HotkeyRecorderView.swift
```diff
@@ -82,6 +82,19 @@ struct HotkeyRecorderView: View {
         fnMonitor = NSEvent.addLocalMonitorForEvents(matching: .flagsChanged) { event in
             guard isRecording else { return event }
 
+            // Accept bare Right Option before the generic Option/chord filter.
+            let chordModifiers: NSEvent.ModifierFlags = [.command, .control, .shift, .function]
+            if event.keyCode == UInt16(HotkeyMonitor.rightOptionKeyCode),
+               event.modifierFlags.contains(.option),
+               event.modifierFlags.intersection(chordModifiers).isEmpty,
+               event.modifierFlags.rawValue & 0x20 == 0 { // NX_DEVICELALTKEYMASK
```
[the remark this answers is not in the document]
fbettag: Replaced the inline bit literal with the same named leftOptionMask constant used by the global handler. The recorder explicitly documents that Caps Lock is allowed, matching the handler and updated PR description (37970e5).

Sources/WisprApp/Services/HotkeyMonitor.swift
```diff
@@ -97,12 +104,12 @@ final class HotkeyMonitor {
...
     /// Registers a global hotkey with the given key code and modifier flags.
     ///
-    /// For keyCode 63 (Fn/Globe) with no modifiers, uses a CGEventTap.
+    /// For bare Fn/Globe or Right Option, uses a CGEventTap.
     /// For all other combinations, uses Carbon's RegisterEventHotKey.
     ///
     /// - Parameters:
     ///   - keyCode: The virtual key code (e.g., 49 for Space, 63 for Fn).
-    ///   - modifiers: Carbon modifier flags (e.g., optionKey = 2048). Use 0 for Fn key.
+    ///   - modifiers: Carbon modifier flags (e.g., optionKey = 2048). Use 0 for modifier-only keys.
```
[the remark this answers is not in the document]
fbettag: Corrected the parameter documentation to name only bare Fn/Globe and Right Option in 37970e5.

Sources/WisprApp/Services/HotkeyMonitor.swift
```diff
@@ -416,6 +428,30 @@ final class HotkeyMonitor {
...
+
+        if !isDown {
+            guard rightOptionDidTrigger else { return false }
+            rightOptionDidTrigger = false
+            onHotkeyUp?()
+            return true
+        }
+
+        let otherModifiers: CGEventFlags = [.maskCommand, .maskControl, .maskShift, .maskSecondaryFn]
+        guard flags.intersection(otherModifiers).isEmpty,
```
sebsto, line 451: **[Correctness — confirming, not a defect]** Nice catch that this handler's `otherModifiers` set deliberately *omits* `.maskAlternate` (unlike `handleFnFlagsChanged`, which includes it), because Right Option *is* an Alternate — and Left Option is excluded separately via `leftOptionMask`. That asymmetry is correct and easy to misread as a bug later. Might be worth one short comment on this line noting *why* `.maskAlternate` is intentionally absent here, so a future reader doesn't "helpfully" add it back and break bare-Right-Option detection. Purely a maintainability nicety.

Sources/WisprApp/Services/HotkeyMonitor.swift
```diff
@@ -36,14 +36,19 @@ final class HotkeyMonitor {
 
     /// Virtual key code for the Fn/Globe key.
     static let fnKeyCode: UInt32 = 63  // kVK_Function
+    static let rightOptionKeyCode: UInt32 = 61  // kVK_RightOption
+
+    /// Device-specific flags distinguish the two Option keys even when both are held.
+    private static let rightOptionMask: UInt64 = 0x40  // NX_DEVICERALTKEYMASK
```
sebsto, line 43: **[Robustness — low, optional]** `rightOptionMask`/`leftOptionMask` (`0x40`/`0x20`) are the device-dependent (`NX_DEVICER/LALTKEYMASK`) bits, which is exactly right for telling the two Option keys apart. These are stable in practice, so no change needed — just flagging that this handler's behaviour depends on the OS continuing to set those device bits in `flagsChanged`. If they were ever absent, `isDown` would stay false and bare Right Option simply wouldn't trigger (fails safe, no false activation), which is the good failure mode. Fine as-is; noting it for the record.

Sources/WisprApp/UI/Settings/HotkeyRecorderView.swift
```diff
@@ -82,6 +82,20 @@ struct HotkeyRecorderView: View {
         fnMonitor = NSEvent.addLocalMonitorForEvents(matching: .flagsChanged) { event in
             guard isRecording else { return event }
 
+            // Accept bare Right Option before the generic Option/chord filter.
+            // Caps Lock is allowed, matching the global hotkey handler.
+            let chordModifiers: NSEvent.ModifierFlags = [.command, .control, .shift, .function]
+            if event.keyCode == UInt16(HotkeyMonitor.rightOptionKeyCode),
```
sebsto, line 89: **[Consistency — low]** This recorder guard nicely mirrors the global handler's filter (chord modifiers + left-Option exclusion), so what you can *record* matches what will *fire*. One tiny divergence for a follow-up if you feel like it: the runtime handler treats Caps Lock as allowed implicitly (it isn't in its `otherModifiers`), while here Caps Lock isn't in `chordModifiers` so it's also allowed — the two agree, good. Keeping the two lists defined next to each other (or sharing one source of truth) would make it obvious they must stay in sync. Not needed for this PR.

sebsto (commented):
Thanks so much for this, Franz — and welcome! 🎉 This is a really nice first contribution: a clear, self-contained feature with a well-written PR description that walks through exactly the edge cases that matter (chord vs. bare key, release-while-another-modifier-held, Caps Lock as a latched state). The event state machine is factored to be testable without a live tap, and the 25 tests cover the tricky transitions directly. I ran `swift test --filter 'RightOptionHotkeyTests|HotkeyMonitorTests'` locally and all 25 pass.

On the security side (the part I looked at hardest, since this touches a global event tap): this is clean. The change reuses the existing Fn tap infrastructure and only widens the routing condition — it does **not** broaden what is captured. `eventsOfInterest` stays `flagsChanged` only, so no keystroke content is ever seen; the handler reads modifier flag bits and nothing else; nothing is logged from the Right Option path and nothing leaves the process. No new privilege or capture surface is introduced. 👍

Everything below is minor / advisory — nothing blocks merge. A couple of the notes are really just me confirming a subtle decision was intentional rather than asking for a change.

Sources/WisprApp/Services/HotkeyMonitor.swift
fbettag: Added the explanation directly above the filter in 729ec52: Right Option itself sets maskAlternate, and Left Option is excluded separately. This is intentional.

Sources/WisprApp/Services/HotkeyMonitor.swift
fbettag: Confirmed: without the device-specific bit, Right Option does not activate. Added that invariant beside the masks in 729ec52. No generic Option fallback is intended because it could mistake Left Option for the dictation key.

Sources/WisprApp/UI/Settings/HotkeyRecorderView.swift
fbettag: Added reciprocal comments linking the recorder and runtime filters in 729ec52, including the shared Caps Lock policy. I kept the AppKit and CoreGraphics flag lists in their respective API types to avoid introducing a conversion abstraction for this PR. This update is comments only; the previously verified 25-test behavior is unchanged. Physical keyboard and GUI verification for this upstream build is still pending.

PR #111, opened 2026-09-17
Report CLI permission denials instead of "No models downloaded"

### LINKED ISSUE #110: wispr-cli reports "No models downloaded" when denied disk access
## Summary

`wispr-cli` reports `No models downloaded. Please open Wispr.app and download at least one model` when the real problem is that the process lacks Full Disk Access. The models are present and working in the GUI. The advice in the message is wrong and unactionable.

## Steps to reproduce

1. Download a model in Wispr.app (GUI shows it as Downloaded/Active, dictation works).
2. From a terminal that has **not** been granted Full Disk Access, run:

```
wispr-cli --list-models
```

Observed:

```
Error: No models downloaded. Please open Wispr.app and download at least one model, then try again. Run --list-models to verify.
```

Expected: an error explaining that the model directory could not be read, and which permission to grant.

## Root cause

The GUI is sandboxed (`ENABLE_APP_SANDBOX = YES`), so models live in its container:

```
~/Library/Containers/com.stormacq.mac.wispr/Data/Library/Application Support/wispr/models/
```

`ModelPaths.base` deliberately points the non-sandboxed CLI at that same path, and **that part works correctly**. This is not a path mismatch.

macOS protects another app's container `Data` directory as app-private data. A non-owning process needs Full Disk Access to read it, and the failure is asymmetric:

- `FileManager.fileExists(atPath:)` uses `stat()` — **permitted**, returns `true`.
- `FileManager.contentsOfDirectory(atPath:)` needs `opendir`/`readdir` — **denied**, returns `EPERM`.

So discovery gets past the existence guard, then both enumeration calls fail. Both are wrapped in `try?`, which discards the error:

```swift
// Sources/WisprCLI/WisprCLI.swift — discoverDownloadedModels()
guard fm.fileExists(atPath: modelsDir.path) else { throw CLIError.noModelsDirectory }

if let variants = try? fm.contentsOfDirectory(atPath: whisperDir.path) { ... }
if let entries  = try? fm.contentsOfDirectory(atPath: modelsDir.path)  { ... }
```

`results` ends up empty, and the caller throws `CLIError.noDownloadedModels`. A permission denial is indistinguishable from an empty directory.

Confirmation from an affected machine:

```
$ ls -al ~/Library/Containers/com.stormacq.mac.wispr        # lists fine
$ ls -al ~/Library/Containers/com.stormacq.mac.wispr/Data
ls: .../Data: Operation not permitted
```

Granting Full Disk Access to the terminal fixes it immediately. This is not SSH-specific — it affects any terminal without the grant, in a GUI session or over SSH.

The same denial hits `ModelPaths.guiDefaultsPlist`, read with `try? Data(contentsOf:)`. A user who fixed only the directory read would next get `CLIError.noActiveModel` — a second misleading message for the same cause.

## Impact

Beyond the wrong message, nothing in the CLI ever prints the path it resolved, and every filesystem error is swallowed. Diagnosing one user report took six rounds of back-and-forth and ruled out several phantom causes (path mismatch, version skew, stale UserDefaults flags, engine state desync) before the permission denial surfaced.

Also note `CLAUDE.md` documents the shared model location as `~/Library/Application Support/wispr/models/`, which is only the fallback path and does not exist on a normal install. That wrong path actively misdirected the investigation. The README does not document the CLI at all.

## Proposed fix

Contained to `Sources/WisprCLI/WisprCLI.swift` plus docs. No change to `ModelPaths` resolution.

- Replace both `try?` enumerations with a single explicit readability probe on `ModelPaths.models`, and reuse its result for the Parakeet scan.
- Add `CLIError.modelsDirectoryUnreadable(path:underlying:)`. Classify permission denials via `NSFileReadNoPermissionError` and `NSPOSIXErrorDomain` `EPERM`/`EACCES` (top-level or underlying). Anything unrecognized is surfaced verbatim rather than replaced with a generic message.
- Emit terminal-appropriate guidance naming the exact remedy: System Settings → Privacy & Security → Full Disk Access, add the terminal app (`/System/Applications/Utilities/Terminal.app`), quit and reopen it. For SSH, add `/usr/libexec/sshd-keygen-wrapper`, and restart any pre-existing `tmux` server with `tmux kill-server`.
- Include the resolved absolute path in the "not set up", "no models downloaded", and read-failure errors.
- Apply the same classification to the preferences plist read, distinguishing "unreadable due to permissions" from "absent".
- Print resolved `ModelPaths.base` to stderr under `--verbose`.
- Fix the documented path in `CLAUDE.md`; add a CLI section to the README stating the Full Disk Access prerequisite.

### Non-goals

The CLI should fail loudly, not work around the permission. No fallback directory, no mirroring models to an unprotected location, no degraded mode.

## Open question, tracked separately

Requiring every user to grant Full Disk Access to their terminal is a poor out-of-box story for a shipped CLI. Dropping the App Sandbox from the GUI would remove this failure mode entirely, and the app ships via Homebrew cask and a notarized pkg rather than the App Store, so the sandbox is not buying distribution. That is a separate distribution and migration decision (existing users would likely need to re-download models) and is not part of this fix.

## Workaround for affected users

System Settings → Privacy & Security → Full Disk Access → `+` → ⌘⇧G → `/System/Applications/Utilities/Terminal.app` (or your terminal). Enable it, then quit and reopen the terminal. Over SSH, add `/usr/libexec/sshd-keygen-wrapper` and run `tmux kill-server` before reconnecting if you use tmux.

Note this is a broad permission: it applies to every command run in that terminal, not just `wispr-cli`.

---

Requirements are written up in `.kiro/specs/cli-model-access-errors/requirements.md`.

sebsto:
Fixes #110.

`wispr-cli` reported `No models downloaded. Please open Wispr.app and download at least one model` to users whose models were present and working in the GUI. The real cause was a missing Full Disk Access grant, and nothing in the output pointed at it.

## Why it happened

The CLI reads models from the GUI's sandbox container, which macOS protects as app-private data. A process that is not the owning app needs Full Disk Access, and the denial is asymmetric:

- `fileExists(atPath:)` uses `stat()` — permitted, returns `true`
- `contentsOfDirectory(atPath:)` needs `readdir` — refused, returns `EPERM`

Discovery got past the existence guard, then both enumerations failed. Both were wrapped in `try?`:

```swift
if let variants = try? fm.contentsOfDirectory(atPath: whisperDir.path) { ... }
if let entries  = try? fm.contentsOfDirectory(atPath: modelsDir.path)  { ... }
```

A permission denial was therefore indistinguishable from an empty directory, and no error named the path it had read. Diagnosing one report took six rounds and ruled out several phantom causes first.

`ModelPaths` resolution was never at fault and is unchanged.

## What changed

**New `FileAccess` in `WisprCore`** — permission classifier, a directory read that throws instead of returning `nil`, and the Full Disk Access guidance text.

It sits in `WisprCore` rather than the CLI for testability: the test target depends on `WisprApp` and `WisprCore`, not on the `WisprCLI` executable target, and adding a dependency on an `@main` target would need changes to both `Package.swift` and the Xcode test target for no design benefit. `CLIError` stays a thin presentation layer.

The classifier matches `NSFileReadNoPermissionError` plus `EPERM` and `EACCES`, at the top level and across `underlyingErrors`. TCC produces `EPERM`, ordinary filesystem permissions produce `EACCES`, and Foundation often wraps the errno in a Cocoa error. A classification miss falls back to `.other` carrying the underlying description, so a miss degrades to a verbose but truthful message rather than a wrong one.

**In the CLI** — one authoritative read of the models directory, reused by the Parakeet scan, so a denial is detected in exactly one place and the directory is no longer read twice. Every discovery error now carries the resolved path. A permission denial on the preferences plist escalates the same way, since it is the same root cause; absent and malformed still mean "active model unknown". `--verbose` prints the resolved models root.

`fileExists` is kept only to tell "never set up" from "set up but unreadable". It cannot serve as a permission check, because it succeeds on protected paths — that is the whole trap.

**Xcode project** — `Sources/WisprCore` uses explicit file references rather than a synchronized group, so `FileAccess.swift` had to be registered or the product build fails to find it. Worth knowing: `swift build` succeeded while `xcodebuild` failed, so SPM alone would not have caught this.

**Docs** — `CLAUDE.md` documented the shared model path as `~/Library/Application Support/wispr/models/`, which is only `ModelPaths.base`'s fallback and does not exist on a normal install. That wrong path actively misdirected the original investigation. The README had no CLI section at all and now documents the tool and its Full Disk Access prerequisite.

## Before and after

Before, with a protected container:

```
Error: No models downloaded. Please open Wispr.app and download at least one model, then try again. Run --list-models to verify.
```

After:

```
Error: Cannot read Wispr's model directory.

  /Users/sst/Library/Containers/com.stormacq.mac.wispr/Data/Library/Application Support/wispr/models

macOS protects one app's data directory from other processes. Grant Full Disk Access to your terminal, then quit and reopen it:

  System Settings › Privacy & Security › Full Disk Access › +
  Add your terminal app, for example:
    /System/Applications/Utilities/Terminal.app

Over SSH, add /usr/libexec/sshd-keygen-wrapper instead. If you use tmux, run "tmux kill-server" and reconnect after granting access, because an existing tmux server keeps its old permissions.
```

## Tests

20 new tests in `wisprTests/FileAccessTests.swift`, across three suites.

- **Classification** — every recognised shape, including a POSIX denial wrapped in a Cocoa error via `NSUnderlyingErrorKey`. Negative cases pin down that `NSFileNoSuchFileError`, `ENOENT`, and a foreign domain carrying the `EPERM` numeric value are not denials.
- **Strict reads** — a readable directory lists contents; an empty directory succeeds rather than throwing, which is the distinction the change rests on; a missing directory throws with a non-permission failure; a directory chmodded to `000` throws `.permissionDenied` rather than returning an empty list. That last test restores the mode in a `defer` so cleanup can descend, and is disabled under `getuid() == 0` since root bypasses directory permissions.
- **Message content** — the denial message names its subject and path and includes the remedy, the Terminal.app path, the quit-and-reopen step, `sshd-keygen-wrapper`, and `tmux kill-server`. One test asserts the denial message never contains the word "download", as a direct regression guard on the original bug. Two more confirm a non-permission failure reports its reason and omits the Full Disk Access advice, so a disk error does not send someone to Privacy & Security.

## Verification

- `swift build`, `xcodebuild -scheme WisprCLI`, and `xcodebuild -scheme wispr` all succeed.
- New tests: 20/20 pass. The mode-`000` case exercised a real `EACCES` rather than being skipped.
- Full suite: 676 tests. One failure, `AudioEngine audio level stream terminates on stop`, which is **pre-existing and unrelated** — it fails identically on the stashed baseline (656 tests, same single failure) and passes when the AudioEngine suite runs in isolation. Timing-sensitive audio hardware test, no audio code touched here.
- Manually exercised the real bug by chmodding the actual container's `models` directory to `000`: the CLI printed the message above with the real path and exit code 1, and after restoring permissions `--list-models` again printed `parakeet-v3  483 MB (active)`. The second half matters — it confirms the strict reads did not introduce a false failure on a working install.

Still worth doing before release: the same check on a machine whose terminal genuinely lacks Full Disk Access, since that produces `EPERM` rather than the `EACCES` a chmod produces.

## Not in scope

The CLI fails loudly rather than working around the permission: no fallback directory, no mirroring models to an unprotected location, no degraded mode.

Requiring every user to grant Full Disk Access to their terminal is still a poor out-of-box story. Dropping the App Sandbox from the GUI would remove this failure mode entirely, and the app ships via Homebrew cask and a notarized pkg rather than the App Store. That is a separate distribution and migration decision — existing users would likely need to re-download models — and is noted as an open question on #110.

## Reviewer note

The `project.pbxproj` diff includes ~7 lines of pre-existing Xcode reordering churn that was already in my working tree before this work, alongside my 4 additions. Happy to strip it if you would rather keep the diff to just the file registration.

Requirements, design, and task breakdown are in `.kiro/specs/cli-model-access-errors/`.


Sources/WisprCore/Utilities/FileAccess.swift
```diff
@@ -0,0 +1,131 @@
...
+    ///   - error: The failure to explain.
+    ///   - what: Subject of the read, e.g. `"Wispr's model directory"`.
+    public static func explain(_ error: FileReadError, what: String) -> String {
+        let header = """
+            Cannot read \(what).
+
+              \(error.path)
+            """
+
+        switch error.failure {
```
sebsto, line 135: **[Maintainability / Low]** `explain` renders the `.other` case as `"The read failed: <reason>"`, but the design doc's *Error message copy* section specifies `"The directory could not be read: <reason>"`. Purely a doc-vs-code wording drift — the code reads fine; the design doc is now stale. Worth a one-line doc tweak so the two don't diverge, non-blocking.

Sources/WisprCore/Utilities/FileAccess.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    private static func isPOSIXDenial(_ error: NSError) -> Bool {
+        guard error.domain == NSPOSIXErrorDomain else { return false }
+        return error.code == Int(EPERM) || error.code == Int(EACCES)
+    }
+
+    /// Maps an arbitrary error onto a `FileReadFailure`.
+    public static func classify(_ error: Error) -> FileReadFailure {
+        isPermissionDenied(error)
+            ? .permissionDenied
```
sebsto, line 102: **[Correctness / Low]** `classify` stores `error.localizedDescription` for the `.other` branch. For a Cocoa error that's the right string; for a bare `NSPOSIXErrorDomain` error `localizedDescription` collapses to a generic "The operation couldn’t be completed. (…)" that drops the errno name. Since permission denials (`EPERM`/`EACCES`) are already peeled off by `isPermissionDenied`, the residual `.other` set is small, so this is minor — but `(error as NSError)` domain+code would be strictly more diagnostic in a bug report. The message stays truthful either way, which is the design's stated bar. Advisory.

sebsto (commented):
**Summary — clean and shippable, nothing blocks merge.**

This fixes #110: the CLI reported `No models downloaded` to users whose models were present but sat behind a Full Disk Access denial, because the directory enumeration was wrapped in `try?` and a permission failure was indistinguishable from an empty directory. The PR replaces both `try?` reads with a strict `FileAccess.readDirectory` that classifies the failure, names the resolved path in every discovery error, escalates a plist-permission denial the same way, and adds `--verbose` root disclosure.

The three things I checked hardest, all good:

- **The core property holds.** A `contentsOfDirectory` failure can no longer reach `.noDownloadedModels`; it becomes `.unreadable`. The single-authoritative-read of the models directory (reused by the Parakeet scan) means a denial is detected in exactly one place. The regression is pinned by a test asserting the denial message never contains the word "download", and by the empty-directory-is-success test that the whole fix rests on.
- **Concurrency and placement follow house style.** `FileReadFailure.other` stores a pre-rendered `String` rather than `any Error`, keeping the type `Sendable` with no `nonisolated(unsafe)` / `Task.detached` escape hatch; the types are `nonisolated` to match `ModelPaths` under `defaultIsolation(MainActor.self)`. Living in `WisprCore` (not the CLI executable target) is the right call for testability.
- **The Xcode trap is handled.** The new `Sources/WisprCore` file uses explicit file references, and `FileAccess.swift` is registered in `project.pbxproj` (build file, file reference, group child, and Sources phase). This is exactly the class where `swift build` passes while `xcodebuild` fails to find the symbol — good that it was caught and that all three CI legs are green.

Everything below is advisory / Low. Tests are thorough (20 across three suites) and idiomatic swift-testing.

Sources/WisprCore/Utilities/FileAccess.swift
```diff
@@ -0,0 +1,131 @@
...
+    /// still surfaces the underlying description, so an unrecognised error
+    /// produces a verbose but truthful message rather than a wrong one.
+    public static func isPermissionDenied(_ error: Error) -> Bool {
+        let nsError = error as NSError
+        if nsError.domain == NSCocoaErrorDomain,
+            nsError.code == NSFileReadNoPermissionError {
+            return true
+        }
+        if isPOSIXDenial(nsError) { return true }
+        return nsError.underlyingErrors.contains { isPOSIXDenial($0 as NSError) }
```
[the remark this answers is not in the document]
sebsto: Implemented, but the stated premise is not the reason — worth separating the two.

**The `NSUnderlyingErrorKey` claim is incorrect.** `NSError.underlyingErrors` already reports the singular key; it is not limited to `NSMultipleUnderlyingErrorsKey`. Verified directly:

```swift
let posix = NSError(domain: NSPOSIXErrorDomain, code: Int(EPERM), userInfo: [:])
let singular = NSError(domain: NSCocoaErrorDomain, code: NSFileReadUnknownError,
                       userInfo: [NSUnderlyingErrorKey: posix])
print(singular.underlyingErrors.count)                                  // 1
print((singular.underlyingErrors[0] as NSError).domain)                 // NSPOSIXErrorDomain
```

The existing test `A POSIX denial wrapped in a Cocoa error is still a denial` constructs exactly that shape and passes on `dc21c08`. It can only pass through the `underlyingErrors` branch, since `NSFileReadUnknownError` fails the Cocoa check and the Cocoa domain fails `isPOSIXDenial`. So the supported shape was never falling through to `.other`.

**The recursion point is correct, for a different reason.** `underlyingErrors` descends exactly one level, which the same probe shows:

```swift
let nested = NSError(domain: NSCocoaErrorDomain, code: NSFileReadUnknownError,
                     userInfo: [NSUnderlyingErrorKey: singular])
print(nested.underlyingErrors.map { ($0 as NSError).domain })           // ["NSCocoaErrorDomain"]
```

A wrapper wrapping a wrapper did hide the denial. Fixed in 2e3c1c4 by walking the chain recursively, with a depth cap so a pathological or self-referential chain cannot recurse without bound.

Three tests added: a denial two wrappers deep, a Cocoa no-permission error nested inside a wrapper, and a 50-deep non-denial chain asserting traversal terminates. Suite is now 23 tests, all passing; full suite 679.

.kiro/specs/cli-model-access-errors/design.md
```diff
@@ -0,0 +1,432 @@
...
+Open Wispr.app and download at least one model, then run --list-models to verify.
+```
+
+The `>` characters stand in for the arrows in the actual strings; the implementation uses the
+Unicode arrow, matching how the guidance reads in the issue.
+
+## Data Models
+
+No new data models. `DownloadedModelInfo` and `TranscribeConfig` are unchanged. The only new type is
+the nested `CLIError.FailureReason` enum described above.
```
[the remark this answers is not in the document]
sebsto: Correct on both counts — implemented in 2e3c1c4.

The Data Models section described a nested `CLIError.FailureReason` that the implementation never had. I moved the types into `WisprCore` during implementation so the classifier and guidance text would be reachable from the test target (`WisprTests` depends on `WisprApp` and `WisprCore`, not on the `WisprCLI` executable target), updated several other sections to match, and missed this one and task 1.1.

Both now describe the real API: `FileReadFailure` (`.permissionDenied` / `.other(String)`) and `FileReadError` (`path` + `failure`) as top-level types in `WisprCore`, with `CLIError.unreadable(what:error:)` carrying a `FileReadError` and `CLIError` gaining no nested types. Data Models is now a table naming all three additions and their module.

Two related staleness fixes while in there, both from the recursion change in the sibling thread:

- The classifier code block in `design.md` still showed the single-level version, and Property 4 still said "among its underlying errors". Both now describe the recursive traversal and the depth cap.
- `requirements.md` Requirement 5 gains a termination criterion as 5.3, which renumbers graceful degradation to 5.4. Updated the two `design.md` cross-references that pointed at 5.3.

Test counts in both documents corrected to 23.

sebsto:
Both review points addressed in 2e3c1c4. Per-point disposition:

**1. Classifier and `NSUnderlyingErrorKey`** (`FileAccess.swift:61`) — **implemented, premise corrected.**

The stated cause is wrong: `NSError.underlyingErrors` does report the singular `NSUnderlyingErrorKey`, so the wrapped shape was never falling through to `.other`. A Cocoa error carrying only that key yields `underlyingErrors.count == 1` with `domain == NSPOSIXErrorDomain`, and the existing test for that shape passes on `dc21c08` — it can only pass via the `underlyingErrors` branch.

The recursion suggestion is nonetheless right, for a different reason: `underlyingErrors` descends exactly one level, so a wrapper wrapping a wrapper did hide the denial. Traversal is now recursive with a depth cap. Evidence and the probe output are in the thread.

**2. Stale spec API descriptions** (`design.md:314`) — **implemented.** The Data Models section and task 1.1 described a nested `CLIError.FailureReason` that never existed; the real types are `FileReadFailure` and `FileReadError`, top-level in `WisprCore`. Also fixed the classifier code block, Property 4, a requirements renumbering, and the test counts. Details in the thread.

Nothing dismissed.

Tests: 3 added, `FileAccess` suite now 23, full suite 679. The single failing test in the full suite, `AudioEngine audio level stream terminates on stop`, is pre-existing and unrelated — it fails identically on the pre-PR baseline and passes when its suite runs in isolation. CI was green on `dc21c08` and is re-running on `2e3c1c4`.

Still open as a reviewer question, unchanged from the description: whether to strip the ~7 lines of pre-existing Xcode reordering churn from the `project.pbxproj` diff.

Sources/WisprCore/Utilities/FileAccess.swift
sebsto: Correct, fixed in b060f7e — doc side, as you suggested.

Kept the code wording deliberately, and noted why in the doc so it does not drift back: `explain` is not directory-specific. It also renders the preferences-file failure, thrown from `guiDefaultsString` as `CLIError.unreadable(what: "Wisprs preferences file", ...)`. So "The directory could not be read" would be actively wrong for that caller, whereas "The read failed" is accurate for both. The subject is already named on the first line by `what`, so nothing is lost by keeping the second line generic.

The design doc now reads:

```
The read failed: <reason>
```

with a line recording that the wording is generic on purpose.

Sources/WisprCore/Utilities/FileAccess.swift
sebsto: Keeping `localizedDescription` — the specific concern does not reproduce. Measured on macOS 27 (26A428):

```
bare POSIX EIO       -> "The operation couldn’t be completed. Input/output error"
bare POSIX ENOTDIR   -> "The operation couldn’t be completed. Not a directory"
bare POSIX ELOOP     -> "The operation couldn’t be completed. Too many levels of symbolic links"
```

Foundation appends the `strerror` text, so the errno is not dropped — it is spelled out in words, which is arguably better than `NSPOSIXErrorDomain 20` in a message aimed at a user rather than a developer.

The premise is also moot for this call path: `contentsOfDirectory` returns Cocoa errors, not bare POSIX ones. Two real failures:

```
missing directory  -> NSCocoaErrorDomain 260, "The folder “here” doesn’t exist."
path is a file     -> NSCocoaErrorDomain 256, "The file “hosts” couldn’t be opened."
```

Both already read well. So `.other` gets a decent string in the shapes that actually occur, and in the hypothetical bare-POSIX shape too. Appending domain+code would add developer-facing noise to every instance of a rare fallback branch to buy diagnostic detail that is already present in prose.

Happy to add it if you would rather have the raw domain+code in bug reports regardless — it is a two-line change and your call as maintainer. Left as-is for now since the stated reason for it does not hold.

Filed nothing as follow-up; if you want it, say so and I will push it into this PR rather than defer it.

Sources/WisprCore/Utilities/FileAccess.swift
```diff
@@ -0,0 +1,150 @@
...
+        if error.domain == NSCocoaErrorDomain,
+            error.code == NSFileReadNoPermissionError {
+            return true
+        }
+        if isPOSIXDenial(error) { return true }
+
+        guard depth < maxUnderlyingErrorDepth else { return false }
+        return error.underlyingErrors.contains {
+            isPermissionDenied($0 as NSError, depth: depth + 1)
+        }
```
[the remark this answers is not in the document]
sebsto: Correct — implemented in ff828fa, and it caught a real inconsistency I introduced.

The cap traded one bug for another: it bounded recursion, but silently reclassified a denial nested below it as `.other`, which contradicts Requirement 5.2 ("anywhere in its chain of underlying errors"). Choosing 5 was an arbitrary constant defending against a case that cannot really happen, at the cost of a case the requirement explicitly covers.

Now terminates on a visited `ObjectIdentifier` set and walks every unvisited underlying error, so no reachable denial is skipped at any depth, and the constant is gone.

Verified the identity assumption the visited set depends on, since it would be unsound if `as NSError` re-boxed:

```swift
let pulled = wrapper.underlyingErrors[0]
ObjectIdentifier(pulled as NSError) == ObjectIdentifier(pulled as NSError)  // true
ObjectIdentifier(pulled as NSError) == ObjectIdentifier(originalPosixError) // true
```

`NSError` is a class and the cast preserves reference identity for errors pulled out of `underlyingErrors`, so the set really does detect revisits rather than accumulating fresh boxes.

Two tests added, one per gap:

- a denial 50 wrappers deep — fails under the old cap, passes now
- a denial reachable only through one branch of an `NSMultipleUnderlyingErrorsKey` chain, since the plural form yields siblings and any of them may hold it

The existing 50-deep non-denial test stays as the termination guard. Suite is 25 tests, full suite 681, `xcodebuild -scheme WisprCLI` green. Requirement 5.3 now also states that nesting depth must not cause a denial to be skipped, so the spec cannot drift back to a cap.

sebsto:
All five review threads resolved. CI green on `ff828fa`.

| # | Comment | Disposition |
|---|---|---|
| 1 | Classifier misses POSIX wrapped via singular `NSUnderlyingErrorKey` | **Implemented, premise corrected** — `underlyingErrors` does report the singular key (measured). The recursion point was right for a different reason: it descends only one level. `2e3c1c4` |
| 2 | Spec describes nonexistent nested `CLIError.FailureReason` | **Implemented** — Data Models and task 1.1 now describe `FileReadFailure` / `FileReadError` in `WisprCore`. `2e3c1c4` |
| 3 | `.other` copy drift: doc says "The directory could not be read" | **Implemented, doc side** — code wording kept, since `explain` also renders the preferences-*file* failure. `b060f7e` |
| 4 | Use domain+code instead of `localizedDescription` for `.other` | **Kept** — Foundation appends `strerror`, so the errno is not dropped; and `contentsOfDirectory` returns Cocoa errors here anyway. Offer standing to add it if you want raw domain+code regardless. |
| 5 | Fixed depth cap misses a denial ≥6 wrappers deep | **Implemented** — cap replaced by a visited-`ObjectIdentifier` set. Caught a real contradiction with my own Requirement 5.2. `ff828fa` |

Net effect of the review: one genuine defect fixed (5), one genuine gap fixed on a false premise (1), two doc/spec corrections (2, 3), one advisory kept with evidence (4).

Tests grew from 20 to 25 in the `FileAccess` suites; full suite 681. Five of the 25 now cover chain traversal specifically, including the two shapes the cap would have missed.

Requirement 5.3 was tightened to say that nesting depth must not cause a denial to be skipped, so the spec cannot drift back toward a cap.

Two items still open for you, neither raised by a reviewer:
- Whether to strip the ~7 lines of pre-existing Xcode reordering churn from the `project.pbxproj` diff.
- Thread 4 above, if you would prefer domain+code in `.other` after all.

